PR #56, opened 2011-11-12
#370: prepare build for uploading artifacts to maven central

jbellmann:
Hi,

one of your 'Fun-Projects' was to provide Voldemort artifacts from maven central. I prepared the build.xml, added the maven-ant-task.jar an pom.xml-template to make it should work.

But there is more to do on your side. Your 'Release-Manager' has to create an Account on Sonatype OSS and to create an 'Project-Request' on Sonatype OSS Jira. You can read the following well written documentation: https://docs.sonatype.org/display/Repository/Sonatype+OSS+Maven+Repository+Usage+Guide

Or you can ask me for some help, if it is possible for me to help. Or if you have any further questions.

This pullrequest relates to:
http://code.google.com/p/project-voldemort/issues/detail?id=370

Cheers,
Jörg


miguel0afd:
+1
It would be great if Voldemort were in Maven Central

FelixGV:
Hi,

I believe this probably won't work as we moved away from the ant build towards gradle instead. In any case, we need to clean up dependencies before Maven Central publishing can be achieved. I've started doing that here: https://github.com/voldemort/voldemort/pull/243

After that, we'll only need to figure out how to depend on bdb-je and then we can publish Voldemort artifacts to Maven Central (:

-F

miguel0afd:
Hi,

in any case, thanks for the work., I'll be around in order to find out when the goal is achieved ;-)

FelixGV:
FYI: All dependencies besides BDB-JE have been cleaned up. We need to test #247 before we are free from all checked in dependencies... We want to be careful because the BDB-JE upgrade might introduce performance regressions.

PR #71, opened 2012-04-07
Project Voldemort maven support

iladriano:
- Exact version on all dependencies (based on lib/jars)

Allows installation of voldemort jar in local repo, setup eclipse, etc.


wenerme:
Any maven coordinate ?

arunthirupathi:
Maven is still on the backlog. But we managed to remove some of the
dependencies but still some of them remains. No one is actively working on
it, if someone from the community can reduce it to just tehuti and tusk,
probably we can get the ball rolling.

Thanks,
Arun.

FelixGV:
Yeah, I think it's going to be a requirement that all of Voldemort's dependencies can be gotten via Maven coordinates before Voldemort itself can be published to Maven Central.

I am going to publish Tehuti to Maven Central pretty soon... I've also fixed the Azkaban dependency in the last release so it's pulled from Maven Central instead of from private-libs.

We need to figure out a way to pull the other dependencies from publicly available Maven repos as well, or to eliminate those other dependencies. Like @arunthirupathi said, this is not high-priority for anyone here, so if the community wishes to help in cleaning up the remaining dependencies, that would be great.

-F

FelixGV:
FYI: The only remaining dependency to pull from Maven before Voldemort can be published to Maven is BDB-JE... tracked in PR #247.

FelixGV:
Oups, closed by mistake.

PR #147, opened 2013-06-11
Made informed stuff optional (off by default) in Voldemor push and pull

sukhendu:
Fixed VoldemortBuildAndPushJob.java to make informed code optional (off by default). Option "voldemort.informed.on" need to be set to true to switch it on.


FelixGV:
This issue is obsolete since the introduction of a mechanism for custom hooks.

PR #224, opened 2014-10-10
config param for Informed logging

kedardoshi:
Put the informed logging code behind a config param


FelixGV:
The Informed code has been removed and replaced with a mechanism for custom hooks.

PR #236, opened 2014-12-30
Fix the Krati entries iterator to properly handle key collisions.

jwlent55:
When using Krati the Fetch command does not always return the expected number of key/value pairs.  The larger the store the more likely that the number of key/value pairs returned would be less than expected.  I tracked this issue down to what I think is a very simple bug in the KratiStorageEngine code - if there are key collisions at an index then the current iterator returned by the call to the entries method only returns the first key at that index.

This update contains a fix for this problem and a new test case that demonstrates the problem.

### DIFF
--- a/contrib/krati/src/java/voldemort/store/krati/KratiStorageEngine.java
+++ b/contrib/krati/src/java/voldemort/store/krati/KratiStorageEngine.java
@@ -105,7 +105,8 @@ public class KratiStorageEngine extends AbstractStorageEngine<ByteArray, byte[],
                 // TODO: Move to DynamicDataStore code
                 ByteBuffer bb = ByteBuffer.wrap(returnedBytes);
                 int cnt = bb.getInt();
-                if(cnt > 0) {
+                // Loop over all keys at this index
+                for(int i = 0; i < cnt; i++) {
                     int keyLen = bb.getInt();
                     byte[] key = new byte[keyLen];
                     bb.get(key);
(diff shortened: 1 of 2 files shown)


arunthirupathi:
I am going to merge this pull request in. But beware that Krati can lead to data corruptions and the storage engine as I know is moot. Stay away from Krati if possible.

jwlent55:
Thanks for the quick review.  When I initially started investigating this issue I noted that work on Krati appeared to have ceased and found some conversations in the voldemort discussion group (like "Krati support?") that suggested Krati may have some issues and was not really a recommended storage engine for Voldemort.  I was therefore not surprised by your suggestion to avoid Krati.

We are looking into alternatives to Krati and BDB and would welcome any suggestions.

In (half hearted) defence of Krati this bug is in the Voldemort wrapper around Krati and at first it looked like a Krati data corruption issue.  It appeared during a routine maintenance procedure used to upgrade several nodes - keys were lost.  Since this bug also impacts the Restore operation one node's data could not be reliably reconstructed from its peer node's data.  It also (I think) impacts the Slop operation if Krati is used for that purpose too.

arunthirupathi:
We use BDB and we are ok with it. We were able to push the cases that used Krati before for list appends to read/modify/write and it works for us. 

What are the limitations you are running into BDB ? For us it worked when we moved some of the high throughput list appends to read the list/appened it/ write it back thought it generated huge amount of garbage and we have to do cluster expansion to handle it.  But this is a solvable problem than the data corruption Krati caused from time to time and fire fighting after that.

arunthirupathi:
What version of BDB JE are you looking at ? BDB je 5.0.88 is what we have and I believe it is GPL version. IANAL ( I am not a lawyer) and TINLA ( This is not legal advice ) if you are running a service you may be fine with BDB JE version included with voldemort. But upgrades to latest version of BDB will not be possible as they moved to AGPL license.

jwlent55:
Thanks for the detailed license information.  Your comments are rekindling an internal debate here.

PR #242, opened 2015-03-13
Further BnP cleanups:

FelixGV:
- Removed Azkaban dependency as much as possible.
- Standardized on using voldemort.utils.Props as much as possible.
- Deleted VoldemortMultiStoreBuildAndPushJob which is not actively used and suffering from code rot.
- Added Content-Length header support in BnP HttpHook.
- Added more utility functions to voldemort.utils.Props
- Safeguards in BnP HttpHook's concurrent code.
- Removed System.exit calls from BnP.


FelixGV:
This has been merged in as part of another commit...

PR #244, opened 2015-03-18
Playing around with rocksdb

bhasudha:
Rebased again.

This is an experimental first cut implementation of Rocksdb storage engine for Voldemort.

It needs better configs management, tuning, lock optimizations, etc.

Includes implementation for APIs such as get, put, delete, getAll and unit tests for testing the StorageEngine and its APIs.

Vinoth, can you take a look at this and let me know?

Thanks,
Sudha


src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -135,6 +136,9 @@
     private String mysqlHost;
     private int mysqlPort;
 
+    private String rdbDataDirectory;
```
vinothchandar: rename to rocksdbDataDirectory? (uniform elsewhere too)

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -567,6 +572,11 @@ public VoldemortConfig(Props props) {
         this.slopPurgeJobMaxKeysScannedPerSec = props.getInt("slop.purgejob.max.keys.scanned.per.sec",
                                                              10000);
 
+        // RocksDB config
+        this.rdbDataDirectory = props.getString("rocksdb.data.dir", "/tmp/rdb_data_dir");
```
vinothchandar: can we derive this from VOLDEMORT_HOME or something by default, like how bdb data directory works?

FelixGV:
I think this is a really great contribution! Thanks for driving this @bhasudha ! And thanks for reviewing @vinothchandar (:

Once we're done with the review and related changes, I suggest squashing it all in one big "Initial RocksDB support" commit. You can include my own commits in the squash, they were pretty minimal anyway and not fully working so I don't see the point of cluttering the git history with them.

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -3231,4 +3241,37 @@ public String getFileFetcherClass() {
...
+    }
+
+    /**
+     * If true, keys will be prefixed by the partition Id on disk. This can
+     * possibly speed up rebalancing, restore operations, at the cost of 2 bytes
+     * of extra storage per key
+     * 
+     * @param rocksdbPrefixKeysWithPartitionId
+     */
+    public void setRocksdbPrefixKeysWithPartitionId(boolean rocksdbPrefixKeysWithPartitionId) {
```
vinothchandar: so it does not make much sense for this flag to even exist now.. (its abundantly clear that the extra 2 bytes in the key is totally worth it, for all the speedup we get with admin ops).. So may be have one server level config to enable it for all store types on the same server?

src/java/voldemort/store/rocksdb/RocksDbStorageConfiguration.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static Logger logger = Logger.getLogger(RocksDbStorageConfiguration.class);
+
+    private final VoldemortConfig voldemortconfig;
+
+    private Map<String, RocksDbStorageEngine> stores = new HashMap<String, RocksDbStorageEngine>();
+
+    public RocksDbStorageConfiguration(VoldemortConfig config) {
+        /**
+         * - TODO 1. number of default locks need to debated. This default is
```
vinothchandar: TODO still relevant? in general we should set the striped lock count to a multiple of number NIO server threads (the idea is there will be only as much locking as number of threads)

src/java/voldemort/store/rocksdb/RocksDbStorageConfiguration.java
```diff
@@ -0,0 +1,115 @@
...
+
+        if(!stores.containsKey(storeName)) {
+            String dataDir = this.voldemortconfig.getRdbDataDirectory() + "/" + storeName;
+
+            new File(dataDir).mkdirs();
+
+            // TODO: Validate those default mandatory options and make them
+            // configurable
+            try {
+                Options rdbOptions = new Options().setCreateIfMissing(true)
```
vinothchandar: can we put in a good default value for rocksdb's block cache?

src/java/voldemort/store/rocksdb/RocksDbStorageEngine.java
```diff
@@ -0,0 +1,490 @@
...
+    public boolean delete(ByteArray key, Version version) throws PersistenceFailureException {
+
+        StoreUtils.assertValidKey(key);
+
+        long startTimeNs = -1;
+
+        if(logger.isTraceEnabled())
+            startTimeNs = System.nanoTime();
+
+        synchronized(this.locks.lockFor(key.get())) {
```
vinothchandar: so for put and delete, we do RMW at our level.. is there a way to push this down to rocksdb (merge operator?)

src/java/voldemort/store/rocksdb/RocksDbStorageEngine.java
```diff
@@ -0,0 +1,490 @@
...
+    public List<Versioned<byte[]>> get(ByteArray key, byte[] transforms)
+            throws PersistenceFailureException {
+        // TODO read locks ?
+        StoreUtils.assertValidKey(key);
+        long startTimeNs = -1;
+
+        if(logger.isTraceEnabled())
+            startTimeNs = System.nanoTime();
+
+        List<Versioned<byte[]>> value = null;
```
vinothchandar: can you add a mode where get() also take a lock based on a config value? (bdb has an internal config for this mode).. Check BdbStorageEngine

test/unit/voldemort/store/rocksdb/RocksdbStorageEngineAPITest.java
```diff
@@ -0,0 +1,276 @@
...
+        } catch(PersistenceFailureException pfe) {
+            Assert.fail("initial put failed unexpectedly. Exception: " + pfe.getMessage());
+        }
+        versionedList.add(value2);
+        return versionedList;
+    }
+
+    @Test
+    public void testVersionedDelete() {
+        System.out.println("*********** testing versioned delete ***********");
```
vinothchandar: get rid of these s.o.plns and replace with logger?

vinothchandar:
I left a few comments.. LGTM overall.. I will wait for a day or two for you to address them.. if you don't find time, I will merge this and make them myself.. keep me posted..

bhasudha:
Thanks @FelixGV ! I will squash them before merging.

Thanks @vinothchandar for the review. I will address the comments on Inday (coming Friday) and merge them.

PR #247, opened 2015-04-06
Upgrade to BDB-JE 5.0.104

FelixGV:
Unit tests pass. No perf test done yet.

# Change log

## Changes in 5.0.104

Corrected the lastKnownUtilization cleaner statistic (EnvironmentStats getLastKnownUtilization). Previously, the stat value reflected the projected utilization if all pending cleaning were completed, i.e., the stat value was always at or above the configured minUtilization setting. Now the stat value reflects the current utilization and will be below the configured minUtilization when cleaning is backlogged or in-progress. \[#23153\] (5.0.104)

Fixed a bug that sometimes resulted in an uncommitted record deletion performed in one transaction to be visible (result in a NOTFOUND result) to an operation performed in another transaction. This bug applies to the use of Database.delete and PrimaryIndex.delete. It does not apply to the use of SecondaryDatabase.delete, SecondaryIndex.delete, or the use of a cursor to perform a deletion. Note that this problem is distinct from a similar bug that was fixed in JE 5.0.98 (\[#22892\]).
\[#23132\] (5.0.104)

## Changes in 5.0.103

Fix cast exception when calling EnvironmentStat.getAvgBatchManual() method. \[#23060\] (5.0.99)

Fixed a bug that could cause an IndexOutOfBoundsException when logging the Environment statistics by the StatLogger. \[#22885\] (5.0.98)

Fixed a bug that could cause an EnvironmentFailureException with the message "DB not found during non-recovery undo/rollback", when a transaction is invalidated in one thread and concurrently used in another thread to perform a write operation. This could only occur with using a transactional, non-replicated database, in a replicated environment. \[#22875\] (5.0.98)

Fixed a bug that could cause the following exception, when an Error is thrown and caught concurrently by two threads performing JE operations.
 Caused by: java.lang.IllegalStateException: Can't overwrite cause
    at java.lang.Throwable.initCause(Throwable.java:320)
    at com.sleepycat.je.dbi.EnvironmentImpl.invalidate(EnvironmentImpl.java:1529)
    ...
\[#22837\] (5.0.98)

Fixed a bug that sometimes resulted in an uncommitted record deletion performed in one transaction to be visible (result in a NOTFOUND result) to an operation performed in another transaction. \[#22892\] (5.0.98)

Added a new statistic for cheaply checking the current utilization of the log, without having to run the DbSpace utility: EnvironmentStats.getLastKnownUtilization. See the javadoc for details. \[#22911\] (5.0.98)

Added a column named time that contains the time the data was logged to the je.config.csv environment statistics file. \[#22919\] (5.0.98)

Fixed a bug that caused immutable Environment configuration properties to not have the correct value in the je.config.csv file. \[#22923\] (5.0.98)

Improved performance when "no sync" durability is used, by performing writes of the buffer outside of the global write lock. This also prevents writing multiple buffers in a single thread, which could cause performance issues on some systems. \[#22988\] (5.0.100)

Fixed a bug that prevented exclusive lock-out of application threads during a preload (via Database.preload or Environment.preload). If applications threads were able to write during the preload, this could cause corruption. \[#23017\] (5.0.101)

The JE cleaner now wakes up whenever a database is removed or truncated. This ensures that disk space is reclaimed as soon as possible, even when there is little or no write activity. \[#22915\] (5.0.101)

CacheMode.EVICT_LN now leaves the parent INs hot after an operation, rather than forcing them to be cold. This allows using EVICT_LN to avoid Java GC problems, while still using an LRU for eviction. \[#23066\] (5.0.101)

Fix DbFilterStat utility to work with the je.config.csv file. Also added envcfg: to the EnvironmentConfig column names in the je.config.csv file. \[#22996\] (5.0.102)

Fix a bug where a NullPointerException was thrown by Database.openCursor when called after closing an invalid Environment. \[#23083\] (5.0.102)

## Changes in 5.0.97

Fixed a bug that causes excessive logging when an environment directory is forcibly removed while the environment is open, during a unit test for example. \[#22625\] (5.0.85)

The internal JE thread for capturing statistics is no longer created when the EnvironmentConfig.STATS_COLLECT config param is false. \[#22659\] (5.0.85)

In an HA application, the probability of hard rollbacks during a master transfer operation has been reduced. \[#22658\] (5.0.85)

Updated the logging and stats file formats to print times in UTC timezone and in a consistent format that includes the full year. The new format is "2013-07-31 13:47:27.753 UTC". \[#22670\] (5.0.85)

Add -outputproperties switch to DbCacheSize. This causes the calculated sizes to be written to System.out in Java properties format. See the DbCacheSize usage command in the javadoc for details. \[#22719\] (5.0.86)

Shutdown HA feeder threads and the replica thread promptly as a result of an EnvironmentFailureException instead of waiting for the environment handle to be closed. The change makes for faster, more robust failover and reduces dependence on application behavior. \[#22637\] (5.0.86)

Eliminates duplicate log messages associated with an InsufficientLogException. The message was logged twice: the first time as a WARNING and again as SEVERE. With this change, the message is logged just once, as a WARNING. \[#22789\] (5.0.87)

Added the EnvironmentConfig.CLEANER_ADJUST_UTILIZATION boolean configuration parameter, to allow disabling of undesired automatic adjustments to log utilization. By default, adjustments are enabled, and the default behavior has not changed. For more information, see the javadoc for this parameter. \[#22768\] (5.0.88)

A performance enhancement was made to improve multi-threaded write performance. The checksum computation was moved out of the critical section during which a global mutex is held. This, together with earlier improvements to the write path, improve write performance overall for multi-threaded applications. However, the performance improvements are limited to standalone JE applications; performance is not improved for HA applications. \[#16925\] (5.0.97)

Fixed a bug that could cause the following exception when an error occurs while reading from the log, and a CRUD operation is in progress. The error while reading invalidates the Environment and no further operations are possible, but the incorrect exception may be misleading.
com.sleepycat.je.EnvironmentFailureException: (JE 5.0.86) Latch not held:
IN1419 UNEXPECTED_STATE: Unexpected internal state, may have side effects.
    at com.sleepycat.je.EnvironmentFailureException.unexpectedState(EnvironmentFailureException.java:390)
    at com.sleepycat.je.latch.SharedLatch.release(SharedLatch.java:181)
    at com.sleepycat.je.tree.IN.releaseLatch(IN.java:551)
    at com.sleepycat.je.tree.Tree.getNextIN(Tree.java:1120)
    at com.sleepycat.je.tree.Tree.getNextBin(Tree.java:935)
    at com.sleepycat.je.dbi.CursorImpl.getNext(CursorImpl.java:1666)
    ...
\[#22772\] (5.0.97)

Fixed a bug that in rare conditions could cause a SecondaryIntegrityException such as the following. This could occur when a given key is being inserted and deleted concurrently in two different threads, and another thread is reading the record using read-uncommitted isolation. This bug is present is all earlier versions of JE 5 and JE 4.x.
Caused by: com.sleepycat.je.SecondaryIntegrityException: (JE 5.0.83) Secondary
refers to a missing key in the primary database
    at com.sleepycat.je.Database.secondaryRefersToMissingPrimaryKey(Database.java:2252)
    at com.sleepycat.je.Cursor.readPrimaryAfterGet(Cursor.java:3882)
    at com.sleepycat.je.SecondaryCursor.readPrimaryAfterGet(SecondaryCursor.java:1519)
    at com.sleepycat.je.SecondaryCursor.retrieveNext(SecondaryCursor.java:1495)
    at com.sleepycat.je.SecondaryCursor.getNext(SecondaryCursor.java:601)
    ...
\[#22603\] (5.0.97)

JE now logs a SEVERE message as follows in the je.info file at startup when -XX:+UseCompressedOops is specified but the heap size is larger than the maximum required value on the JVM platform for using this option. For example, on the Oracle JVM, -Xmx29g or smaller must be specified to use this option. On non-Oracle JVM platforms (more specifically those JVM platforms not including the com.sun.management.HotSpotDiagnosticMXBean class), JE does not currently detect this situation and does not log a message.
2013-08-29 19:41:45.965 UTC SEVERE \[.\] -XX:+UseCompressedOops was specified but
is not in effect, probably because the heap size is too large for this JVM
option on this platform.  This is likely to cause an OutOfMemoryError!
Note that in this situation, and presumably on all JVMs, the JVM itself also outputs a warning message at startup such as the following:
OpenJDK 64-Bit Server VM warning: Max heap size too large for Compressed Oops
\[#22657\] (5.0.97)

Fixed problems related to log cleaning and the estimation of log utilization. Previously, in one reported case log utilization was adjusted incorrectly (see EnvironmentConfig.CLEANER_ADJUST_UTILIZATION), resulting in very small values for the LNSizeCorrectionFactor (EnvironmentStats.getLNSizeCorrectionFactor()). In addition, repeated probing of the same log file was observed. Both of these problems have been fixed. Thanks to Vinoth for reporting this on OTN and working with us to find the problem. \[#22814\] (5.0.97)

Change Learner Agent to only respond to a query for a master when it's in the Master or Replica state, to ensure that the master information is current. Prior to this change, obsolete information could prevent the replication group from recovering on it's own and electing a new master in some rare circumstances. \[#22815\] (5.0.97)

Avoid throwing an InsufficientLogException when a Replication Node that has been inactive for a period exceeding ReplicationConfig.REP_STREAM_TIMEOUT, but is otherwise reasonably current, rejoins the group. Such Replication Nodes can now rejoin the group without going through a Network Restore operation. \[#22782\] (5.0.97)

Fixed a bug that could result in a stack trace like the one below:
Caused by: java.lang.IllegalArgumentException: toKey out of range
    at java.util.TreeMap$AscendingSubMap.headMap(TreeMap.java:1719)
    at java.util.TreeSet.headSet(TreeSet.java:338)
    at java.util.TreeSet.headSet(TreeSet.java:372)
    at com.sleepycat.je.cleaner.Cleaner.deleteSafeToDeleteFiles(Cleaner.java:739)
    at com.sleepycat.je.cleaner.Cleaner.updateFilesAtCheckpointEnd(Cleaner.java:935)
    at com.sleepycat.je.recovery.Checkpointer.doCheckpoint(Checkpointer.java:797)
    at com.sleepycat.je.recovery.Checkpointer.onWakeup(Checkpointer.java:508)
    at com.sleepycat.je.utilint.DaemonThread.run(DaemonThread.java:148)
This problem could occur in rare circumstances when using the DbBackUp or DiskOrderedCursor APIs. \[#22822\] (5.0.97)

Added EnvironmentConfig.ENV_RECOVERY_FORCE_NEW_FILE, which is to be used after performing a restore from backup to force creation of a new log file prior to recovery. For more information, see the javadoc for this parameter and the "Restoring from a backup" section in the DbBackup javadoc. \[#22834\] (5.0.97)


FelixGV:
FYI, this is the final step before being able to publish Voldemort to Maven Central, since this eliminates the last dependency checked into the private-lib directory.

vinothchandar:
Please canary this in production for a couple of weeks, before merging.. We have had some cases of innocuous seeming changes affect us unexpectedly.

FelixGV:
Oh yeah of course. We're not planning to merge this in until it's fully certified.

FYI: This is not a high-priority item for us ATM.

busbey:
Is there a current timeline on merging this in? YCSB is preparing for its first release in ~3 years and I'd like to update the Voldemort version off of the current super old version (0.81), but we only add things that are accessible on *some* maven repo (does not need to be Maven Central).

arunthirupathi:
@busbey  Can you please help us in getting this merged, These are the open items at our end to merge this change

1) 5.0.88 and 5.0.104 binary compatbility. Create db using one version, open it using other version and vice versa.
2) Is there any fix that is left out of 5.0.104 ? 5.0.88 is custom built for Voldemort by BDB JE and is never published to maven central.
3) Any other information you can collect about 5.0.104.

This will help us in fixing this last issue before we make it publishable to maven central.

busbey:
I'm afraid I don't have the slack time to do much on evaluating the changes between those two versions; it would take me too long even to come up to speed on what they do.

If you are interested in binary compatibility, I'd suggest you take a look at the [Java API Compliance Checker](http://ispras.linuxbase.org/index.php/Java_API_Compliance_Checker). It can create a report that compares the two, specifically in the context of how voldemort uses it. Let me know if you run into problems with the tool and I can help on that.

arunthirupathi:
API compatibility is easier to check and the build with 104 is already passed. create the database using the 5.0.88 , write some records. Read them using 5.0.104 and vice versa. 

Is there any volunteer from the open source community to help us with this ? We are very busy at this time and without the community involvements, driving this to completion is going to take time.

In any case, you are using the oldest version which has tons of bugs. The new code has tons of fixes and improvements. So you should make a case for the new version regardless.

vinothchandar:
Usually, the BDB folks are very good at calling out backwards compatibility issues, if any. 
The only thing that can make this move faster is actual production canary and making sure there are n't regressions. This needs a very large scale deployment such as at Linkedin. So, if you guys can quickly frankenwar this on a bunch of boxes, that would be useful.. 

That said, yes @busbey comparing against older voldemort releases is pointless. http://www.slideshare.net/vinothchandar/voldemort-prototype-to-production-nectar-edits is a part of the a linkedin tech talk, which goes over a lot of the newer re-implementations.. 
Let me know if I can help in anyway.

busbey:
> API compatibility is easier to check and the build with 104 is already passed. create the database using 
> the 5.0.88 , write some records. Read them using 5.0.104 and vice versa.

The Java API Compliance Checker also checks binary compatibility. The only thing it can't tell you is if there are behavior changes. Shouldn't the existing automated tests already tell you if there is a behavior change?

>Is there any volunteer from the open source community to help us with this ? We are very busy at this 
> time and without the community involvements, driving this to completion is going to take time.
>
> In any case, you are using the oldest version which has tons of bugs. The new code has tons of fixes
> and improvements. So you should make a case for the new version regardless.

I can empathize with a busy schedule and the conflicting needs of internal deadlines and the open source community. Sometimes these things just take a while to line up.

Over in YCSB we've been doing a lot of work to bring the community back from the dead and a big part of that is easing the process of getting up and working with the code base. Right now the lack of availability in a Maven repo is a blocker for us updating our Voldemort dependency; most of the users of YCSB won't be familiar with Voldemort and we can't ask them to download a different project and install it locally as a part of our build process.

That said, I also agree that it's a bad idea to ship out a known obsolete version of the Voldemort client. That's very likely to lead to folks getting an inaccurately negative view of the project, presuming they could even use it with a modern server-side deployment. Maybe it would be best if we just disable Voldemort support for this upcoming release.

What do you think? We're scheduled to have monthly releases, so adding back in support once y'all have the time to publish artifacts should be straight  forward.

FelixGV:
Hi @busbey,

Monthly re-runs sound great! We'll let you know once we have a properly Mavenized artifact for Voldemort, but in the mean time, I guess YCSB will probably be Voldemort-less.

Thanks for reaching out.

-F

PR #249, opened 2015-04-07
Client v2 shared buffer

arunthirupathi:
This is a rewrite of the Voldemort Hot Path and Client NIO.

1) Client does 50-80% less allocation on the hot path. ( They share read and write buffer, isCompleteResponse does not allocate the objects, just moves the pointer past and while sending data it allocates required number of bytes). When I ran the client with this fix, the total amount of memory allocated by my program was less than 50% and the used heap dropped by more than 50%.
2) Similar fixes for the Server. Client read and write buffers are combined into one.
3) Client NIO does not block anymore.

After these fixes the 95th and 99th percentiles were a constant. 

I am sending this pull request so that the open source community can review it.

I am in the process of splitting the change into client and server and release them independently.

### DIFF
--- a/src/java/voldemort/server/niosocket/AsyncRequestHandler.java
+++ b/src/java/voldemort/server/niosocket/AsyncRequestHandler.java
@@ -121,6 +121,16 @@ public class AsyncRequestHandler extends SelectorManagerWorker {
     }
 
     @Override
+    protected void connect(SelectionKey selectionKey) throws IOException {
+        throw new IOException("Not implemented for Server sockets");
+    }
+
+    @Override
+    protected void reportException(IOException e) {
+        // Not handled for Server side.
+    }
+
+    @Override
     protected void read(SelectionKey selectionKey) throws IOException {
         int count = 0;
 
--- a/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
+++ b/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
@@ -54,6 +54,7 @@ public class ClientRequestExecutor extends SelectorManagerWorker {
 
     private long expiration;
     private long startTime;
+    private long timeoutMs;
     private boolean isExpired;
     protected ByteBufferContainer bufferContainer;
 
@@ -104,14 +105,8 @@ public class ClientRequestExecutor extends SelectorManagerWorker {
         return false;
     }
 
-    public synchronized void addClientRequest(ClientRequest<?> clientRequest,
-                                              long timeoutMs,
-                                              long elapsedNs) {
-        if(logger.isTraceEnabled())
-            logger.trace("Associating client with " + socketChannel.socket());
-
-        this.clientRequest = clientRequest;
-
+    private void computeExpirationTime(long timeoutMs, long elapsedNs) {
+        this.timeoutMs = timeoutMs;
         startTime = System.nanoTime();
 
         if(elapsedNs > (Time.NS_PER_MS * timeoutMs)) {
@@ -129,7 +124,22 @@ public class ClientRequestExecutor extends SelectorManagerWorker {
             throw new IllegalArgumentException(errorMessage);
         }
 
-        resetStreams();
+    }
+
+    public void setConnectRequest(ClientRequest<?> clientRequest, long timeoutMs) {
+        this.clientRequest = clientRequest;
+        computeExpirationTime(timeoutMs, 0);
+    }
+
+    public synchronized void addClientRequest(ClientRequest<?> clientRequest,
+                                              long timeoutMs,
+                                              long elapsedNs) {
+        if(logger.isTraceEnabled())
+            logger.trace("Associating client with " + socketChannel.socket());
+
+        this.clientRequest = clientRequest;
+        computeExpirationTime(timeoutMs, elapsedNs);
+        outputStream.getBuffer().clear();
 
         boolean wasSuccessful = clientRequest.formatRequest(outputStream);
         outputStream.getBuffer().flip();
@@ -237,12 +247,45 @@ public class ClientRequestExecutor extends SelectorManagerWorker {
             if(logger.isTraceEnabled())
                 logger.trace("Finished read for " + socketChannel.socket());
 
+            resetStreams();
             selectionKey.interestOps(0);
         }
         ClientRequest<?> originalRequest = completeClientRequest();
 
         if(originalRequest == null && logger.isEnabledFor(Level.WARN))
             logger.warn("No client associated with " + socketChannel.socket());
+    }
+
+    @Override
+    protected void reportException(IOException e) {
+        clientRequest.reportException(e);
+    }
+
+    @Override
+    protected void connect(SelectionKey selectionKey) throws IOException {
+        if(!checkTimeout())
+            return;
+
+        if(socketChannel.finishConnect() == false) {
+            return;
+        }
+
+        if(logger.isDebugEnabled()) {
+            // check buffer sizes you often don't get out what you put in!
+            if(socketChannel.socket().getReceiveBufferSize() != this.socketBufferSize) {
+                logger.debug("Requested socket receive buffer size was " + this.socketBufferSize
+                             + " bytes but actual size is "
+                             + socketChannel.socket().getReceiveBufferSize() + " bytes.");
+            }
+
+            if(socketChannel.socket().getSendBufferSize() != this.socketBufferSize) {
+                logger.debug("Requested socket send buffer size was " + this.socketBufferSize
+                             + " bytes but actual size is "
+                             + socketChannel.socket().getSendBufferSize() + " bytes.");
+            }
+        }
+
+        addClientRequest(clientRequest, timeoutMs, 0);
     }
 
     @Override
--- a/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java
+++ b/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java
@@ -16,8 +16,8 @@
 
 package voldemort.store.socket.clientrequest;
 
-import java.net.ConnectException;
 import java.net.InetSocketAddress;
+import java.nio.channels.ClosedChannelException;
 import java.nio.channels.ClosedSelectorException;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
@@ -31,17 +31,18 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
 
 import voldemort.common.nio.SelectorManager;
+import voldemort.store.nonblockingstore.NonblockingStoreCallback;
 import voldemort.store.socket.SocketDestination;
 import voldemort.store.stats.ClientSocketStats;
 import voldemort.utils.DaemonThreadFactory;
 import voldemort.utils.Time;
+import voldemort.utils.pool.KeyedResourcePool;
 import voldemort.utils.pool.ResourceFactory;
 
 /**
@@ -71,7 +72,8 @@ public class ClientRequestExecutorFactory implements
                                         int socketBufferSize,
                                         boolean socketKeepAlive,
                                         ClientSocketStats stats,
-                                        String identifier) {
+                                        String identifier,
+                                        ClientRequestExecutorPool executorPool) {
         this.connectTimeoutMs = connectTimeoutMs;
         this.soTimeoutMs = soTimeoutMs;
         this.created = new AtomicInteger(0);
@@ -105,7 +107,7 @@ public class ClientRequestExecutorFactory implements
     /**
      * Close the ClientRequestExecutor.
      */
-
+    @Override
     public void destroy(SocketDestination dest, ClientRequestExecutor clientRequestExecutor)
             throws Exception {
         clientRequestExecutor.close();
@@ -124,8 +126,10 @@ public class ClientRequestExecutorFactory implements
      * 
      * @param dest {@link SocketDestination}
      */
-
-    public ClientRequestExecutor create(SocketDestination dest) throws Exception {
+    @Override
+    public void createAsync(final SocketDestination dest,
+                            final KeyedResourcePool<SocketDestination, ClientRequestExecutor> pool)
+            throws Exception {
         int numCreated = created.incrementAndGet();
         if(logger.isDebugEnabled())
             logger.debug("Creating socket " + numCreated + " for " + dest.getHost() + ":"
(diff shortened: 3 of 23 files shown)


vinothchandar:
Given how large the change is, can you explain your actual change in more detail? (Both the problem and solution), so its easier to review.

arunthirupathi:
Vinoth,
        I believe it will be easier if you review commit by commit. I
believe I have given enough context in each of the commit description.

On a high level these are the changes I did.

1) Share the read/write buffer for Client and Server ( Admin handler is
left untouched as it has around 30 operations). Our client sends the
complete request, waits for the complete response. Server receives the
complete request and sends the complete response. So only read or write
buffer is used at any given point in time. So I combined both of these
buffers. This saves upto 1 GB on some of our existing production clusters.
2) NIO reads in chunks. So to verify if you have the complete request( in
server) / complete response ( in client) it has to be parsed. This parsing
does not need to create objects. The code previously created byte arrays
and discarded them and returned true or false. Now that is avoided and it
is done by Advancing the pointers inside the buffer.
3) While Writing things to the outputStream if the buffer does not have
enough space, the buffer space is doubled unconditionally even for the last
write. So if the buffer is 64 KB and if you want to write a put request in
client for 256 KB you will always allocate 512 KB buffer. Now the sizes are
pre-calculated so it will grow and shrink once. If Server needs to write
GetAll response, and if it crosses teh buffer size, it was creating garbage
on every value put that crosses teh boundary.
4) Some more in-efficiencies in the extra garbage allocation.

On one of the internal tools I ran ( have a store with long Key and byte
array value. I limited the keys from 1 to 10 million and generated a random
byte array from 30 to 45 KB for each put. I did 240 gets, 60 getalls with 5
random keys and 60 puts). The new code allocated less than 50% of the
entire space for the entire app. Of course the scenario I did had very less
app code and the memory is entirely dominated by the voldemort library. In
real world this will greatly vary.

The new code allocates 80% less on the hot path.  Because of the fix 1, the
stable size is also reduced which gives it more room to work with and less
garbage collections. Because of the issues 2-4, young gen garbage
collections are less frequent.

NIO is a problem that Plagued Voldemort for a very long time. We have an
internal high throughput customer and if one of the nodes go down, randomly
other nodes are marked down. This is because of a design flaw with
Selector. Selector is supposed to be continuously calling Select in a tight
loop. But it can try to create connections some time. The connection
creation is not asynchronous and blocks the thread first for establshing
the connection (Remember the fibonacci back off
https://github.com/voldemort/voldemort/blob/master/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java#L155
it has Thread.sleep inside of it)  and
once for protocol negotiation.
https://github.com/voldemort/voldemort/blob/master/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java#L209

Now since the selector is blocked, other connections that are queued for
read and write incurs extra latency and timeouts and causes random node
failures. I have rewritten the NIO layer not to do blocking connections
anymore. Now since the connection creation is asynchronous, the exceptions
during connection creations also need to be handed down to the requests so
that failure detector can correctly take the nodes down. This is little
risky change, but it is required and I have baked it for a good amount of
time. I am trying to get this tested with more number of partners, before
merging the NIO change.

The buffer change is relatively less risky and I solidifed the unit tests
to add the binary protocol verification ( send byte array and expect byte
array to make sure that I am not modifying the client and server protocol).
https://github.com/voldemort/voldemort/commit/224ce69f962cad96986c94011b2d3ac1a0755701
I also added more test cases for the request and response,
https://github.com/voldemort/voldemort/commit/d004dd863636f2b21d8f2beb52c763cca5ed5449,
so I am confident on the allocation changes.

Client Sharing Read/Write Buffer
On a high level, this uses the same buffer for Client read and Write.
https://github.com/voldemort/voldemort/commit/c112c8b572cee477944c814c54af0217e0ce7bc8?w=1

Seperate the Admin and Client Request Handler factory
https://github.com/voldemort/voldemort/commit/68067c30f5f4b9a50ba9fc1a6bc2f06e5f381dde?w=1

 Use Shared Buffer for Server Client Request Handler
https://github.com/voldemort/voldemort/commit/866cf86556b5414de526eea896cd179ceee92047?w=1


Eliminate One extra allocation during Put
https://github.com/voldemort/voldemort/commit/b17558a7ec2874d76f92a6fc821dff3d85de822f?w=1


Stop allocation on isCompleteResponse
https://github.com/voldemort/voldemort/commit/8fe08a7169b9b6f8a15a4727139111e72f8fce9b?w=1

NIO style connect
https://github.com/voldemort/voldemort/commit/056e442ca6cc74ac4d92f9af131eee50589fe852?w=1

Please let me know if you need further information.

Thanks,
Arun.

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
```diff
@@ -116,7 +116,27 @@ public void writeGetRequest(DataOutputStream outputStream,
     }
 
     public boolean isCompleteGetResponse(ByteBuffer buffer) {
-        return isCompleteResponse(buffer, VoldemortOpCode.GET_OP_CODE);
+        try {
+            DataInputStream inputStream = new DataInputStream(new ByteBufferBackedInputStream(buffer));
+            checkException(inputStream);
+            if(!skipeResults(inputStream, buffer)) {
```
bhasudha: 'skipeResults'  -  Did you meant skipResults?

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
```diff
@@ -191,6 +229,28 @@ public boolean isCompleteGetAllResponse(ByteBuffer buffer) {
...
+                                 ByteArray key,
+                                 byte[] value,
+                                 byte[] transforms,
+                                 VectorClock version,
+                                 RequestRoutingType routingType) {
+        int size = 1 + ByteUtils.getUTFMaxLength(storeName) + 1;
+        if(protocolVersion > 1) {
+            size += 1;
+        }
+        size += (4 + key.length());
```
bhasudha: Hi Arun, can you change these hard coded numbers into constants ?

src/java/voldemort/server/protocol/vold/ClientRequestHandler.java
```diff
@@ -0,0 +1,138 @@
+package voldemort.server.protocol.vold;
```
bhasudha: Can we change the package name to 
"package voldemort.server.protocol.native"  instead of "
"package voldemort.server.protocol.vold" ?

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
arunthirupathi: Nice catch will fix it, but in a different change list, after the NIO code is also merged.

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
arunthirupathi: My bad will fix it

src/java/voldemort/server/protocol/vold/ClientRequestHandler.java
arunthirupathi: This will cause lots of change and lots of merge conflict, I have too much code in motion. So would avoid this one. The protocol is still named as Voldemort-native in most of the code, so vold seems better to me than the native.

src/java/voldemort/client/protocol/RequestFormat.java
```diff
@@ -68,6 +70,17 @@ public void writeGetAllRequest(DataOutputStream output,
     public Map<ByteArray, List<Versioned<byte[]>>> readGetAllResponse(DataInputStream stream)
             throws IOException;
 
+    /*
+     * returns the max expected put request size, An implementation can avoid it
+     * by returning the value SIZE_UNKNOWN
+     */
+    public int getExpectedPutRequestSize(String storeName,
```
vinothchandar: I think you do this for only put because, thats the only place where we don't know if we are pulling out a rabbit or elephant from the storage engine..

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
```diff
@@ -116,7 +116,27 @@ public void writeGetRequest(DataOutputStream outputStream,
...
+                return false;
+            }
+        } catch(VoldemortException e) {
+        } catch(IOException e) {
+            return false;
+        }
+        return !buffer.hasRemaining();
+    }
+
+    private boolean skipeResults(DataInputStream inputStream, ByteBuffer buffer) throws IOException {
```
vinothchandar: can you please add javadocs here? I get the idea that you are trying to just wait enough for the  'len' bytes to be available and just skip over.. but someone has to reconstruct this state everytime.. (on an unrelated note, @bhasudha can you throw the doc you had on VNP protocol formats out into an external wiki page? No one really knows that the wire format it unless they read code)

src/java/voldemort/common/nio/ByteBufferContainer.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    public void ensureSpace(int writeLen) {
+        int need = (writeLen - this.buffer.remaining());
+
+        if(need <= 0) {
+            return;
+        }
+
+        growBuffer(buffer.capacity() + need);
```
vinothchandar: this is the meat of your fix.. the *2 is gone.. Sweet!  Overall, I think this would work fairly well.. Using just one buffer is the right thing to do, since any database needs the entire key to do work :P ..

vinothchandar: again javadocs please

src/java/voldemort/server/niosocket/AsyncRequestHandler.java
```diff
@@ -66,9 +71,63 @@ public AsyncRequestHandler(Selector selector,
...
+        }
+    }
+
+    /**
+     * Flips the output buffer, and lets the Selector know we're ready to write.
+     * 
+     * @param selectionKey
+     */
+
+    protected void prepForWrite(SelectionKey selectionKey) {
```
vinothchandar: why protected? I thought you were just moving this around from SelectorManager

src/java/voldemort/server/niosocket/AsyncRequestHandler.java
```diff
@@ -187,10 +247,7 @@ protected void write(SelectionKey selectionKey) throws IOException {
 
         // If we don't have anything else to write, that means we're done with
         // the request! So clear the buffers (resizing if necessary).
-        if(outputStream.getBuffer().capacity() >= resizeThreshold)
-            outputStream.setBuffer(ByteBuffer.allocate(socketBufferSize));
-        else
-            outputStream.getBuffer().clear();
+        outputStream.clear();
```
vinothchandar: So, this path will shrink the buffer in case there was a request greater than the upperBound, correct?  So, while getting rid of the second buffer is a win in terms of memory footprint (still BDB is going dominate the heap nonetheless anyways).. This growing and shrinking is the root cause of the mythical comm buffer problem. It boils down to the upperSizeBound is incorrect and we need to do some heap size checking to see if we can tolerate this value of not.. (eg:  say lowerbound is 64kb and higher is 128kb and request is 129kb, when you clear(), the bufferContainer(),reset() will reset this back to 64kb right?  (Am I getting this correct?).. Imagine this happening in a loop with lots of connections.. that problem still remains I think..

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
```diff
@@ -235,6 +257,38 @@ protected void read(SelectionKey selectionKey) throws IOException {
     }
 
     @Override
+    protected void reportException(IOException e) {
+        clientRequest.reportException(e);
+    }
+
+    @Override
+    protected void connect(SelectionKey selectionKey) throws IOException {
```
vinothchandar: so you are moving the connect() effectively into NIO as well, instead of blocking the thread.. That seems to be essence of it? (If so +1, again one more thing that was always on the backburner.. super happy to see this getting traction).. 

NIO selector does not scale super linearly to number of connections, based on some experiments I ran a long time back. Just something to keep in mind, as pending connections piling up.. Client side may not be that much of deal technically.. pls let me know how it goes in prod..

src/java/voldemort/utils/pool/KeyedResourcePool.java
```diff
@@ -123,35 +126,53 @@ public V checkout(K key) throws Exception {
...
+            long totalNonBlockingElapsedNs = 0;
+            long iterStartTime = 0;
+            long totalBlockingElapsedNs = 0;
+            final long MAX_WAIT_TIME = 300 * Time.NS_PER_MS;
 
-            if(resource == null) {
-                long nonBlockingElapsedNs = System.nanoTime() - startNs;
-                long timeRemainingNs = resourcePoolConfig.getTimeout(TimeUnit.NANOSECONDS)
-                                       - nonBlockingElapsedNs;
+            while(resource == null && (iterStartTime = System.nanoTime()) < endNs) {
```
vinothchandar: Looks like you spin until you get a connection checked out? Won't you lock up the selector manager threads.. (Thats one of the reasons dr.j went this route)

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java
```diff
@@ -206,26 +165,32 @@ public ClientRequestExecutor create(SocketDestination dest) throws Exception {
...
+            };
+
+            NonblockingStoreCallbackClientRequest<String> clientRequest = new NonblockingStoreCallbackClientRequest<String>(pool,
+                                                                                                                            dest,
+                                                                                                                            protocolRequest,
+                                                                                                                            clientRequestExecutor,
+                                                                                                                            callback,
+                                                                                                                            stats);
+
+            clientRequestExecutor.setConnectRequest(clientRequest, timeoutMs);
```
vinothchandar: So, in effect, when there is a connection creation, you do the roundrobin to figure out which selector manager needs to handle the connection request.. and then when it comes back with a socket, does the actual socket get registered to something else or the same selector manager.. (Just for my understanding, it should not matter either ways, as long as we are preserving the round robin)

vinothchandar:
+ 1 overall. Left some comments/clarifications.

src/java/voldemort/client/protocol/RequestFormat.java
arunthirupathi: The Serve still has no clue whether it is pulling out a rabbit or elphant :) . That will require a protcol change. But currently the problem is let us say if you write a value that is 65KB. The client will allocate a ByteBuffer for 130KB ( if it exceeds the size, it doubles the buffer of the required size). And after that it shrinks it to 64KB again. With this fix, the client will just grow the buffer to 65KB and leave it there. I went after all the low hanging fruits , but the Protocol is left intact.

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
arunthirupathi: Vinoth, I will add the javadocs comments. Sorry for not commenting it. All of the work here is done in my personal time and I made a case why we need these fixes which was fortunately approved. So I could not spend enough time on it as I would like it. But that should not be an excuse and I will fix it in a separate commit.

src/java/voldemort/common/nio/ByteBufferContainer.java
arunthirupathi: Yep, all of our client and Server reads the input buffer completely and then writes it out. With this fix, I have cleaned up the Client and the Server.  But I haven't touched the Admin Server code. The tricky thing on the server side is reading and writing happens inside the same method, for Admin it meant touching 32 different functions, so I have skipped that for now.

src/java/voldemort/server/niosocket/AsyncRequestHandler.java
arunthirupathi: Copy/Paste mistake, it was Protected inside SelectorManager, it can be private now. Will fix it.

src/java/voldemort/server/niosocket/AsyncRequestHandler.java
arunthirupathi: The comm buffer issue should be alleviated now. The Admin Buffer is still a mess, and I am working on fixing it, expect a pull request in another month or so.

You are also right about BDB. On read-write on our use cases the overall space savings for the heap size is 4%. But this will let the servers double the number of connections being handled without increasing the memory foot print. The space savings on read only is 20-25% as they run with a very small java heap though.

Your case on growing and dropping back is also exact, the only way to avoid that now is modify the config so that you know your 95th percentile so that growing and shrinking does not happen on each request. The other things you can do is pre-allocate a buffer bool and you can do scatter/gather and return the additional buffers at the end back to the pool. Ideally I believe the connections should not have a buffer. They should be just connections and when they get request they should grab a buffer from the pool. Currently Voldemort can only take a certain number of connections because of the buffer associated with the connections. If we make the buffer association with the connection dynamic the Server can handle few 100,000 connections without issues.

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
arunthirupathi: There are 2 main issues with the Selector today. One is Connect blocks. The second is Protocol Negotiation blocks as well. https://github.com/voldemort/voldemort/blob/master/src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java#L209

Selector should be running in a tight loop reading/writing from sockets. If Selector blocks during connect/protocol negotiation, though the server could have responded the client requests can't receive them as Selector is blocked. This is the root cause of the Search issue, we investigated for more than 2-3 years. On Search, when a node was dying it accepted connect, but never sent the ACK, so the selector blocked and all the other requests queued on the same selector timed out. So we were marking random nodes down. 

On experimental runs in our staging environment for zoned clusters the patch, keeps the 95th percentile at 3 ms(min 2, max 5) and 99th at 5ms.  Unpatched current clients have 95th at 4ms( min 2, max 100) and 99th 10ms( min 4, max 700). This patch lets you have the predictability and after this fix Selector will never block.

So on a high level, this lets you achieve 2 things.
1) Handle the failure scenarios well on a non-zoned cluster. This was always a problem today for us, if a node goes down most of our clients need bouncing.
2) On a Zoned cluster, this gives you a great latency improvements.

src/java/voldemort/utils/pool/KeyedResourcePool.java
arunthirupathi: This is not a spin lock definitely. I try to wait 2/3 times instead of one big wait. Now the connection creation is asynchronous so if one thread triggers the async and waits, other thread could have snatched before the first thread could call wait. So instead of one big wait, now the thread tries to break the time into 2 or 3 chunks and try to do that. Spin lock is not an option and it will burn the CPU like hell. git is not letting me expand the code here, I believe the number is 2 and not 3, but not sure.

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutorFactory.java
arunthirupathi: The actual socket gets registered to the same selector manager. Please look at the line below selectorManager.add(clientRequestExecutor); This registers the socket to the same Selector Manager.

arunthirupathi:
@vinothchandar  Thanks for your comments, much appreciated. I will address all your comments but in a separate commit, is that OK with you. I have 3 branches with subset of these commits and canary-ing them. Addressing the comments before merging it to master is going to take much more time for me and most of this work is not budgeted work. But will address them as soon as the commits are merged into the master.

src/java/voldemort/server/niosocket/AsyncRequestHandler.java
FelixGV: Nice suggestions for future work...

Using heuristics about 95th percentile request size would be pretty interesting and shouldn't be too hard. The buffers pool is probably not such a low hanging fruit, but also interesting for later.

FelixGV:
Nice discussion @vinothchandar and @arunthirupathi, thanks for the insights.

singhsiddharth:
@arunthirupathi  Let's budget some time in the coming sprint(s) so that you address the review comments.

vinothchandar:
@arunthirupathi yeah go ahead.. The direction you are headed is totally valid. Thanks for doing this

src/java/voldemort/client/protocol/vold/VoldemortNativeClientRequestFormat.java
bhasudha: @vinothchandar  yes sure!

arunthirupathi:
All code changes in this branch is merged. Because of rebasing the commit id differed. Have addressed the review comments mostly in this branch https://github.com/arunthirupathi/voldemort/commits/primaryPartitionFetch , Hopefully will send another big review in next 2 months. This time, I am completely re-writing the Streaming APIs for Voldemort.

PR #253, opened 2015-04-23
Ro compression

bhasudha:
Enable compression on BnP jobs and Voldemort RO Server.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/KeyValueWriter.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/KeyValueWriter.java
@@ -36,6 +36,8 @@ public interface KeyValueWriter<K, V> {
     public static final String CHECKSUM_FILE_EXTENSION = ".checksum";
     public static final String GZIP_FILE_EXTENSION = ".gz";
 
+    public static final int DEFAULT_BUFFER_SIZE = 65536;
+
     public static enum CollisionCounter {
 
         NUM_COLLISIONS,
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ It is used at LinkedIn by numerous critical services powering a large portion of
 cd ~/workspace
 git clone https://github.com/voldemort/voldemort.git
 cd voldemort
-ant release
+./gradlew clean jar
 ```
 
 ### Start Server ###
--- a/build.xml
+++ b/build.xml
@@ -1,576 +1,70 @@
 <?xml version="1.0"?>
 
-<project name="voldemort" basedir="." default="all">
-  <property file="gradle.properties" />
+<project name="voldemort" basedir="." default="bail-out">
 
-  <property name="name" value="voldemort" />
-  <property name="display.name" value="Voldemort" />
-  <property name="author" value="Jay Kreps, Roshan Sumbaly, Alex Feinberg, Bhupesh Bansal, Lei Gao, Chinmay Soman, Vinoth Chandar, Zhongjie Wu" />
-  <property environment="env" />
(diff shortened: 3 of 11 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/KeyValueWriter.java
```diff
@@ -25,6 +25,13 @@
 // Interface used by reducers to layout the datqa on disk
 public interface KeyValueWriter<K, V> {
 
+    public static final String COMPRESSION_CODEC = "GZIP";
```
FelixGV: It might be cleaner to put all these compression-related constants in a parameterized enum, though I don't think that's a blocker for merging. Let's just maybe add a TODO so we keep it in mind if/when we want to add support for other compression types?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
```diff
@@ -316,10 +321,48 @@ else if (build && push && dataDirs.size() != 1) {
         }
         if ((!build && push) || (build && !push)) {
             log.warn("DEPRECATED : Creating one build job and separate push job is a deprecated strategy. Instead create"
-                    + "just one job with both build and push set as true and pass a list of cluster urls.");
+                     + "just one job with both build and push set as true and pass a list of cluster urls.");
         }
     }
 
+    private String getMatchingServerSupportedCompressionCodec(int nodeId) {
+        AdminClient adminclient = new AdminClient(clusterUrl.get(0),
```
FelixGV: This code only checks the first cluster. This kind of makes sense given that we build only once for all clusters. That being said, it might be good to be prudent here and make this whole function loop over all the clusters? It would then return the compression type only if all clusters support it... This problem can potentially be avoided by careful operations, so if you'd rather avoid adding the extra check, perhaps it would make sense pull in an SRE to ask them if that's ok...?

singhsiddharth: I thought about this issue too but then figured that since we keep a cluster symmetric across all colos, this should be ok. That is, we shouldn't ever be in a situation where ro cluster in one colo is running a server version with compression enabled, while the same cluster in a different colo isn't.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/KeyValueWriter.java
bhasudha: Done.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
FelixGV: Assuming we stop BnP for all colos until all colos are upgraded, then that's ok...

bhasudha: We tend to keep the clusters in symmetry. So dint bother about checking all clusters. Anyway I will ask the SREs on what they think.

FelixGV:
@bhasudha, this is already merged in right?

bhasudha:
yes Felix!

-Sudha

FelixGV:
All right, cool, just wanted to make sure. I'll close the PR then (:

PR #261, opened 2015-05-15
Output stats file for RO files download

arunthirupathi:
.stats directory will be created and will contain last X (default: 50)
stats file.

If a version-X is fetched a file with the same name as this directory
name will contain the stats for this download.

The stats file will contain the individual file name, time it took to download
and few other information.

Added unit tests for the HdfsCopyStatsTest

Sample Output of one file.

Starting fetch at 1431673167472MS from webhdfs://blah/node-0
Time, FileName, StartTime(MS), Size, TimeTaken(MS), Attempts , TotalBytes
2015-05-15 06:59:27.567,.metadata,1431673167509,86,58,0,86
2015-05-15 06:59:27.628,145_1_0.data,1431673167588,39,40,0,39
2015-05-15 06:59:27.663,413_0_0.data,1431673167628,37,35,0,37
2015-05-15 06:59:27.712,145_1_0.index,1431673167663,12,49,0,12
2015-05-15 06:59:27.757,413_0_0.index,1431673167712,12,45,0,12
2015-05-15 06:59:27.759, Completed at 1431673167759MS. Total bytes Copied 186 . Expected Total bytes 186 . Time taken(MS) 287

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
@@ -1,0 +1,226 @@
+package voldemort.store.readonly.fetcher;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.Date;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.log4j.Logger;
+
+import voldemort.annotations.jmx.JmxGetter;
+import voldemort.utils.Time;
+
+public class HdfsCopyStats {
+
+    private final String sourceFile;
+    private final long startTimeMS;
+    private volatile long bytesSinceLastReport;
+    private volatile long totalBytesCopied;
+    private volatile long lastReportNs;
+    private volatile long totalBytes;
+
+    private BufferedWriter statsFileWriter = null;
+    public static final String STATS_DIRECTORY = ".stats";
+    private final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+    private static final Logger logger = Logger.getLogger(HdfsCopyStats.class);
+
+    private static void deleteExtraStatsFiles(File statsDirectory, int maxStatsFile) {
+        if(maxStatsFile <= 0) {
+            return;
+        }
+
+        try {
+            File[] files = statsDirectory.listFiles();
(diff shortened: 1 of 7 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
```diff
@@ -0,0 +1,226 @@
...
+                                     long timeTakenMS,
+                                     int attempts,
+                                     long totalBytesDownloaded) {
+        reportStats(file.getName() + "," + startTimeMS + "," + fileSize + "," + timeTakenMS + ","
+                    + attempts + "," + totalBytesDownloaded);
+    }
+
+    public void complete() {
+        long nowMS = System.currentTimeMillis() ;
+        reportStats(" Completed at " + nowMS + "MS. Total bytes Copied " + totalBytesCopied
```
FelixGV: Human-readable time would be nice. If you want to also include epoch time for easier parsing by scripts, I'm fine with it... as you wish.

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -355,6 +356,8 @@ public VoldemortConfig(Props props) {
                                                      VoldemortConfig.DEFAULT_KERBEROS_REALM);
         this.fileFetcherClass = props.getString("file.fetcher.class",
                                                 VoldemortConfig.DEFAULT_FILE_FETCHER_CLASS);
+        this.readOnlyStatsFileEnabled = props.getBoolean("readonly.stats.file.enabled", true);
+        this.readOnlyMaxVersionsStatsFile = props.getInt("readonly.stats.file.max.versions", 50);
```
FelixGV: I think we can have a much bigger default here... These files should be very small, so I don't think there's a lot of value in aggressively pruning them. How about making the default 1000? For stores that push every 4 hours, 1000 pushes will yield less than 6 months of stats. For stores that push once a day, it's basically going to last "forever". That sounds reasonable to me.

We can always bump it back down if we have stats disk footprint problems. What do you think?

arunthirupathi: For a directory with 5 files it is about 350 bytes + 500 bytes for header/footer. If the directory has 150 files it is going to 10KB. If we keep 50 stats files that is going to 500 MB per store. For a cluster with 150 stores that is already 75 GB and which is 7.5% on some of our production clusters. So I am worried about increasing this number beyond 50. Please let me know if this number makes sense ?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
arunthirupathi: Will do.

PR #262, opened 2015-06-03
Added logging to detect hung async jobs in AdminClient.waitForCompletion

FelixGV:


src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -869,11 +869,25 @@ public String waitForCompletion(int nodeId,
 
             String description = null;
             String oldStatus = "";
+
+            // State to detect hung async jobs
+            long oldStatusTime = -1;
+            long timeSinceLastWarning = -1;
```
arunthirupathi: should this be called lastStatusReportTime

FelixGV: Hmm you're right. I chose the name hastily and it is misleading. How about ```lastHungStatusWarnTime``` ?

arunthirupathi: The variable is updated whenever status is reported, or warning is reported. So I am not sure the warn should be in the name. This is  a very minor detail, so I would leave it to you which way you want to go with.

FelixGV: I see what you mean. ```lastStatusReportTime``` makes sense, I'll go with that.

FelixGV:
Superseded by #263. Closing.

PR #263, opened 2015-06-03
Added SO_TIMEOUT config (default 30 mins) in ConfigurableSocketFactory

FelixGV:
Also added logging to detect hung async jobs in AdminClient.waitForCompletion


FelixGV:
Hi @arunthirupathi,

I amended the commit again to include your suggestion of wiring in the SO_TIMEOUT setting through ```VoldemortConfig```. Unfortunately, upon further investigation, it seems like this setting won't be picked up when using WebHDFS, but I think we should still merge this in as the extra logging in AdminClient.waitForCompletion is valuable nonetheless...

Let me know if you'd like to see any further changes. Otherwise, feel free to merge in.

-F

PR #266, opened 2015-06-06
ConnectionException is not catastrophic

arunthirupathi:
I believe this will be the last fix in making the Voldemort Client deals with Server failures in a robust 
manner.

The 2 main fixes are for, clear the queue when a node is marked down ( currently they are left to timeout).
When the process hits the zombie state, deal with it as if the process is dead to recover from those errors faster.

1) If a connection timesout or fails during protocol negotiation,
they are treated as normal errors instead of catastrophic errors.
Connection timeout was a regression from NIO connect fix. Protocol
negotiation timeout is a new change to detect the failed servers
faster.
2) When a node is marked down, the outstanding queued requests are
not failed and let them go through the connection creation cycle.
When there is no outstanding requests they can wait infinitely until
the next request comes up.
3) UnreachableStoreException is sometimes double wrapped. This causes
the catastrophic errors to be not detected accurately. Created an utility
method, when you are not sure if the thrown exception could be
UnreachableStoreException use this method, which handles this case
correctly.
4) In non-blocking connect if the DNS does not resolve the Java throws
UnresolvedAddressException instead of UnknownHostException. Probably an
issue in java. Also UnresolvedAddressException is not derived from IOException
but from IllegalArgumentException which is weird. Fixed the code to handle
this.
5) Tuned the remembered exceptions timeout to twice the connection timeout.
Previously it was hardcoded to 3 seconds, which was too aggressive when the
connection for some use cases where set to more than 5 seconds.

Added unit tests to verify all the above cases.

### DIFF
--- a/src/java/voldemort/store/UnreachableStoreException.java
+++ b/src/java/voldemort/store/UnreachableStoreException.java
@@ -36,4 +36,11 @@ public class UnreachableStoreException extends VoldemortException {
         super(s, t);
     }
 
+    public static UnreachableStoreException wrap(String s, Throwable t) {
+        if(t instanceof UnreachableStoreException) {
+            return new UnreachableStoreException(s, t.getCause());
+        }
+        return new UnreachableStoreException(s, t);
+    }
+
 }
--- a/src/java/voldemort/client/protocol/admin/SocketPool.java
+++ b/src/java/voldemort/client/protocol/admin/SocketPool.java
@@ -90,7 +90,7 @@ public class SocketPool {
 
             return sas;
         } catch(Exception e) {
-            throw new UnreachableStoreException("Failure while checking out socket for "
+            throw UnreachableStoreException.wrap("Failure while checking out socket for "
                                                 + destination + ": ", e);
         }
     }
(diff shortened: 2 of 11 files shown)


src/java/voldemort/store/UnreachableStoreException.java
```diff
@@ -36,4 +36,11 @@ public UnreachableStoreException(String s, Throwable t) {
         super(s, t);
     }
 
+    public static UnreachableStoreException wrap(String s, Throwable t) {
```
FelixGV: It seems a bit dirty to me that the ```ThresholdFailureDetector``` relies on an exception's cause for some of its behavior... The cause should be purely informative, IMHO, not for controlling behavior. I'm not sure there is a guarantee provided that an exception's cause will preserved, since stacktraces have no such guarantee when the code JITted.

Regardless, assuming that relying on exception causes is actually clean, this wrap function is problematic, since it clips part of the exception. The part of the stack between the ```new UnreachableStoreException``` on line 41 and the cause should be omitted, which will make debugging very confusing...

It would probably be cleaner to keep all chained exceptions and recurse through them in order to find catastrophic ones. This would at least allow us to preserve full stacktraces for debugging purposes.

arunthirupathi:
Merged.

PR #274, opened 2015-06-22
Avro 1.7.7

kevinsu:


FelixGV:
Hi @kevinsu,

I've been meaning to look into upgrading Avro for a little while, and I think you're going down the right path here, but I'm afraid this can't be merged in as is.

As far as Build and Push is concerned, it's pretty well isolated from the rest of the Voldemort stack, and the wire format is the same between Avro 1.4.x and 1.7.x, so upgrading BnP should not be a big deal.

However, we have many Voldemort clients stuck on Avro 1.4 who need the Voldemort client to continue binary compatible with Avro 1.4, and that is unfortunately impossible since the APIs change between Avro 1.4 and 1.7.

If only BnP was upgraded to 1.7, that could be fine, but there is actually little incentive in doing that... *

The more interesting part would be to have "hybrid" serialization code which can use reflection at runtime to figure out whether the Avro 1.4 or 1.7 APIs are available in the classpath, and then use the appropriate implementation accordingly.

* The only incentive to running BnP with Avro 1.7 is that Hadoop 2 pulls Avro 1.7 by default in its classpath. In our case, we actually solve that problem by shadowing BnP's avro dependency down to 1.4...

-F

arunthirupathi:
I explored this in the past. Kevin, the core of the problem is the
BinaryEncoder non backward compatible changes. Other changes in Build and
Push, Admin Command need not be backward compatible as they are tools and
we can ask people to upgrade.

When I explored it in the past, there is only class common between 1.4 and
1.7 and that is BlockingBinaryEncoder .

Does it make sense for you to explore can we use this class and achieve the
same functionality, which will make Voldemort work seamlessly with both 1.4
and 1.7 ? You need to use the only APIs that are common in both of these
releases.

I don't even know whether it provides the same functionality, performance
or is it too hacky to use that class. So an investigation and update on
that would be super helpful.

Thanks,
Arun.

kevinsu:
Will look at getting the 1.4 + 1.7 hybrid ness going, though that seems very far away.  Will reopen when I have a better idea of what to do going forward.

PR #275, opened 2015-06-28
Ro server refactor

arunthirupathi:
Refactored the fetching of Hdfs Into a separate class, so that it can be overriden easily in the future. This code is reusable if we want to alter the file layout from Node to Partition or use Primary fetch within Voldemort. 

If 2 or more pushes are going on currently, they will produce totally erroneous results on the Bnp log because the Status object is set at HdfsFetcher level. Now it is an additional parameter. StoreName, PushVersion, Metadatastore ( for getting cluster and node id) information are the additional parameters passed in but currently not used.

Tested the code in isolation, by downloading a directory using the hdfs-fetcher script and it downloaded and validated the checksum correctly.

I want this refactor to be included in the next BnP work for throttling before the code deviates more. Hdfs and Hadoop tests passed, other tests are currently being run.


FelixGV:
Hi @arunthirupathi,

Can you elaborate on why (or when) would the BnP logs contain erroneous results? Are you saying this is currently an already existing problem or is it a regression introduced by this PR?

Thanks.

-F

arunthirupathi:
In the Current BnP , there is only one fetcher object for all fetches which
is reused. The AsyncOperationStatus is set on the fetcher before the fetch
is called. Let us say there are two pushes P1 is going on and push P2 was
started little later. Since there is a single fetcher will hold only P2
status.

So P1 will not notice any update in BnP logs. P2 will see both P1 and P2
updates on BnP logs. Does it make sense ?

Look at this line
https://github.com/voldemort/voldemort/blob/master/src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java#L1100

Thanks,
Arun.

FelixGV:
Yeah, got it. Thanks for the explanation.

-F

PR #277, opened 2015-06-30
Improved path handling and validation in VoldemortSwapJob

FelixGV:
BnP doesn't work against a local file-system without this change.


singhsiddharth:
lgtm

FelixGV:
Thanks @singhsiddharth. I actually did some more testing and tweaked this further (amended my commit).

PR #279, opened 2015-07-15
Rewrite of the EventThrottler code to use Tehuti.

FelixGV:
- Makes throttling less vulnerable to spiky traffic sneaking in "between the interval".
- Also fixes throttling for the HdfsFetcher when compression is enabled.


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
```diff
@@ -64,21 +63,17 @@ public BasicFetchStrategy(HdfsFetcher fetcher, FileSystem fs, HdfsCopyStats stat
...
-     * @param stats
-     *            Stats for measuring the transfer progress
      * @param checkSumType
      *            Type of the Checksum to be computed for this file
      * @return A Checksum (generator) of type checkSumType which contains the
      *         computed checksum of the copied file
      * @throws IOException
      */
-    private CheckSum copyFileWithCheckSum(HdfsFile source, File dest, CheckSumType checkSumType) throws IOException {
+    private CheckSum copyFileWithCheckSum(HdfsFile source, File dest, CheckSumType checkSumType) throws Throwable {
```
arunthirupathi: Can you make it throw IOException, instead of throwable, which is very broad.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
```diff
@@ -89,82 +84,86 @@ private CheckSum copyFileWithCheckSum(HdfsFile source, File dest, CheckSumType c
         OutputStream output = null;
         long startTimeMS = System.currentTimeMillis();
 
-        for (int attempt = 0; attempt < fetcher.getMaxAttempts(); attempt++) {
```
arunthirupathi: I believe your settings remove the space between the for, while and if with single space. This introduced lots of change. I generally prefer it with the spaces between them.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/ThrottledInputStream.java
```diff
@@ -0,0 +1,49 @@
...
+        if (throttler != null && read >= 0) {
+            throttler.maybeThrottle(1);
+        }
+        return read;
+    }
+
+    @Override
+    public int read(byte b[], int off, int len) throws IOException {
+        int read = in.read(b, off, len);
+        stats.recordBytesTransferred(read);
```
arunthirupathi: last read returns -1, so should this call be made only if bytes read is not -1

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/ThrottledInputStream.java
```diff
@@ -0,0 +1,49 @@
...
+                                   HdfsCopyStats stats) {
+        super(in);
+        this.throttler = throttler;
+        this.stats = stats;
+    }
+
+    @Override
+    public int read() throws IOException {
+        int read = in.read();
+        stats.recordBytesTransferred(1);
```
arunthirupathi: look at the comment below.

arunthirupathi:
Left some comments, overall looks good. I will have lots of merge conflict with you on some of the white spaces and refactorings ( renames). I will rebase once you merge it in.

But we should follow the common white space pattern to not run into this in the future.

https://github.com/arunthirupathi/voldemort/commits/replicaFromPeer

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
FelixGV: I agree with your preference, it's prettier with a space. However, the culprit is not my editor ;) ...

The original code was "the ugly format", which you changed to "the pretty format" as part of your code move in the ```FetchStrategy``` refactoring here:

https://github.com/voldemort/voldemort/commit/d70ed85e51620620b7ad2f21ca71f5e3537d01e8#diff-8645ea9a3775f1da51779f7d7d12d8e2L489

When I rebased my code on top of master, I copied the whole body of the function rather than editing each line one by one, so I brought back the old formatting, which I (and my editor) hadn't changed.

Anyway, I'll remove that change to keep your new format before merging into master.

-F

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
FelixGV: Again, I agree with you that dealing with ```Throwable```s is generally a bad idea, but this is how the original code was, before your recent refactoring:

https://github.com/voldemort/voldemort/commit/168cb69f3732ac7164f2e66ae8f1ac7239e62323#diff-bed2a573ebc48c26a9eacc4cccd7ef90R180

So my change is actually bringing back the ```Throwable``` that was already there, rather than introducing it in the first place.

I will need to search our production logs to investigate what is it that we actually catch in this block. If some of the HDFS APIs we use actually throw ```Throwable```s and we rely on that for retrying, then we may, unfortunately, need to keep that in there.

-F

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/ThrottledInputStream.java
FelixGV: That's true, I'll change that.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
arunthirupathi: Sorry for the trouble caused here, from time to time my eclipse dies and I have to start the editor formatting from the first. It happens once every 6 months or so, this happened recently to my macbook and hence the code I wrote from macbook are mostly busted. Now I have fixed the formatting.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
arunthirupathi: Sorry for the trouble caused here, from time to time my eclipse dies and I have to start the editor formatting from the first. It happens once every 6 months or so, this happened recently to my macbook and hence the code I wrote from macbook are mostly busted. Now I have fixed the formatting.

arunthirupathi: You can ignore the formatting, I will fix them later.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
FelixGV: Ok, I verified the logs and I think it's safe to change the signatures and catches to ```IOException```, like you suggested.

I should merge this into master soon.

-F

FelixGV:
I merged this in in a squashed commit addressing all the concerns raised.

arunthirupathi:
Thanks Felix appreciate that.

PR #282, opened 2015-07-21
Some unit test improvements.

FelixGV:
- Increased max attempts to spin a Voldemort server in tests from 3 to 10.
- Made Voldemort's Time implement Tehuti's Time for easier inter-op.
- Fixed a bogus test in DataCleanupJobTest (it was actually failing silently without any assertions).


arunthirupathi:
Was it merged in and just the review needs to be closed ?

FelixGV:
I think this change cannot be merged as is. The increased number of attempts for spinning up test servers is already addressed by another commit, IIRC. There is also the discussion on the thread interruption which remains open. It would be worthwhile to drive this to completion so that we get the fix to the ```DataCleanupJobTest```.

PR #283, opened 2015-07-21
Fix --is-avro option handling

elad:


elad:
This file is deprecated.

PR #289, opened 2015-07-31
Upgraded to Hadoop 2.3.0-cdh5.1.5 and cleaned up the Kerberos authentication code.

FelixGV:


singhsiddharth:
LGTM

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -215,85 +215,40 @@ private Configuration getConfiguration(String sourceFileUrl, String hadoopConfig
...
-                        isValidFilesystem = true;
-                    } catch(InterruptedException e) {
-                        logger.error(e.getMessage(), e);
-                    } catch(Exception e) {
-                        logger.error("Got an exception while getting the filesystem object: ");
-                        logger.error("Exception class : " + e.getClass());
-                        e.printStackTrace();
-                        for(StackTraceElement et: e.getStackTrace()) {
-                            logger.error(et.toString());
+        // TODO: Remove this for loop if we don't see the errors anymore.
```
arunthirupathi: This code is not thread safe. Previously this was done on a synchronized this block.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -215,85 +215,40 @@ private Configuration getConfiguration(String sourceFileUrl, String hadoopConfig
...
                 }
 
-                if(isValidFilesystem) {
-                    break;
-                } else if(attempt < maxAttempts - 1) {
-                    logger.error("Attempt#" + attempt
-                                 + " Could not get a valid Filesystem object. Trying again in "
-                                 + retryDelayMs + " ms");
-                    sleepForRetryDelayMs();
+                fs = path.getFileSystem(config);
```
arunthirupathi: Is the getFileSystem call thread safe ? Previously this was in a synchronized block, now it is out of the synchronization block

FelixGV: The comments in our code said doAs() was not thread-safe which sometimes resulted in NPEs being thrown. We do not need to do doAs() anymore (that was a bug in the old WebHDFS implementation) so I think the point is moot. However, if getFilesystem() throws any exception, it'll be caught and retried.

arunthirupathi:
LGTM

FelixGV:
FYI, this is not ready yet. I left it running over the weekend on my test cluster and the token did not get auto-renewed correctly as I expected. I read the Hadoop source and I think I know what's going on. I added some more code in my latest commit to address the issue. We'll see how it goes over the next few days...

src/java/voldemort/utils/ExceptionUtils.java
```diff
@@ -0,0 +1,15 @@
+package voldemort.utils;
+
+/**
+ * Static utility functions to deal with exceptions.
+ */
+public class ExceptionUtils {
+    public static boolean recursiveClassEquals(Throwable throwableToInspect, Class throwableToLookFor) {
```
FelixGV: @arunthirupathi, check this out.

I think this is the type of stuff we would have needed when you were dealing with the client-side exception getting wrapped once too many times.

It seems to me like it's rarely a good idea to look at the cause of an exception, but if you're gonna do it, it usually makes sense to recurse all the way into the call stack, rather than expect you'll always find what you need exactly n levels deep.

If you agree, can we change the error handling code you had worked on to use this new utility function? This way, we'd get un-modified stacktraces in our logs, and make the code a bit cleaner.

-F

arunthirupathi: I agree that this is better way to do it, but I am little scary of touching the old code. My mindset for making change is , make it, test it in EI, simulate in EI . At this point, I will not touch that code for now , but if some of us touch this in the future, we should fix it.

arunthirupathi:
LGTM, you can merge after testing is done.

FelixGV:
Yeah. I'll let it bake for 2-3 days. Then if there aren't any problems, I'll squash and merge.

Thanks for the review!

FelixGV:
This has been running fine in my test cluster for several days.

I squashed and merged.

PR #295, opened 2015-08-11
Fixes to data cleanup job test

FelixGV:
This includes a trivial fix for a broken test. Another commit also changes a default internal setting of the EventThrottler so that it makes more sense in the general case.

### DIFF
--- a/src/java/voldemort/utils/EventThrottler.java
+++ b/src/java/voldemort/utils/EventThrottler.java
@@ -43,7 +43,7 @@ import java.util.concurrent.TimeUnit;
 public class EventThrottler {
 
     private static final Logger logger = Logger.getLogger(EventThrottler.class);
-    private static final long DEFAULT_CHECK_INTERVAL_MS = 50;
+    private static final long DEFAULT_CHECK_INTERVAL_MS = 1000;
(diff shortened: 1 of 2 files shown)


singhsiddharth:
Can you explain in brief why you are bumping up the number ?

FelixGV:
Yes sure.

I originally set the ```EventThrottler```'s default window size to 50 ms, because the previous implementation was checking rates at that interval. I think the main reason why the previous implementation's window was so small is that it didn't have rolling windows so it needed to be small in order to avoid overshooting too much.

However, such small interval does not work well with Tehuti rates when the target throttling rate is less than 20/sec. The reason being that it will register one event, then measure its rate over the span of 50 ms, and come up with a rate of 20 events / seconds. If the throttler code was set to target 1 QPS, it would then determine it needs to sleep for 19 seconds, even if that event was the only one that would come during that second...

There could be use cases where one truly wants to throttle with a 50 ms interval, but I don't think that's necessary for us. Since it breaks the low throughput case, it seems like an inconvenient default. We can always pass in whatever interval we want in order to override the default in specific use cases, if need be.

-F

singhsiddharth:
Makes sense. Thanks for the explanation.

PR #292, opened 2015-08-04
Make BnP work on Hadoop 2.x

elad:
Also let it work standalone (without Azkaban).

Combines stuff from #274 and https://github.com/FelixGV/voldemort/commit/e224b96376104f9a0960259d295c8c8346a39e0e

PR #297, opened 2015-08-12
Bnp runner script so it works more easily without Azkaban

FelixGV:
An incremental change on top of @elad's PR #292.

A BnP fat jar can be generated with ```./gradlew bnpJar```, which can then be used via ```./bin/run-bnp.sh```.

-F


config/readonly_two_nodes_cluster/node0/config/server.properties
```diff
@@ -8,23 +8,15 @@ max.threads=100
...
-mysql.port=1521
-mysql.user=root
-mysql.password=3306
-mysql.database=test
-
 #NIO connector settings.
 enable.nio.connector=true
 
 request.format=vp3
-storage.configs=voldemort.store.bdb.BdbStorageConfiguration, voldemort.store.readonly.ReadOnlyStorageConfiguration
```
singhsiddharth: Any idea why BDB was put in here in the first place ?

FelixGV: Some copy-pasta, I guess.

FelixGV:
This is merged in. Thanks for the contrib @elad !

Issue #278, opened 2015-07-01
 Kerberos authentication is mandatory for Build and Push 


jeremiq:
Kerberos authentication is currently required and not configurable. This makes it difficult to test  build and push jobs in any unsecured cluster. 

With all other configurations in place, build and push will fail like this: 

```
30-06-2015 17:35:20 PDT build_and_push_avro INFO - Caused by: voldemort.VoldemortException: Failed while waiting for async task (Fetch store) at node 0 to finish
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at voldemort.client.protocol.admin.AdminClient$RPCOperations.waitForCompletion(AdminClient.java:925)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at voldemort.client.protocol.admin.AdminClient$RPCOperations.waitForCompletion(AdminClient.java:951)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at voldemort.client.protocol.admin.AdminClient$ReadOnlySpecificOperations.fetchStore(AdminClient.java:3752)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at voldemort.store.readonly.swapper.AdminStoreSwapper$1.call(AdminStoreSwapper.java:150)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at voldemort.store.readonly.swapper.AdminStoreSwapper$1.call(AdminStoreSwapper.java:145)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - 	at java.lang.Thread.run(Thread.java:745)
30-06-2015 17:35:20 PDT build_and_push_avro INFO - Caused by: voldemort.VoldemortException: voldemort.VoldemortException: File fetcher failed for file:/tmp/localhost/node-0 and store 'avro-custom-example-1' Reason: 
[...]

PR #298, opened 2015-08-12
Allow RO servers to run without Kerberos enabled.

FelixGV:
In order to address Issue #278.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -196,12 +196,17 @@ public class HdfsFetcher implements FileFetcher {
 
             String security = config.get(CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION);
 
-            if(security == null || !security.equals("kerberos")) {
-                logger.error("Security isn't turned on in the conf: "
-                             + CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION + " = "
-                             + config.get(CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION));
-                logger.error("Please make sure that the Hadoop config directory path is valid.");
-                throw new VoldemortException("Error in getting Hadoop filesystem. Invalid Hadoop config directory path.");
+            if (security == null) {
+                throw new VoldemortException(CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION +
+                        " is undefined in the config. It must be set to either 'kerberos' or 'simple'.");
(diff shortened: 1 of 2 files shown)


singhsiddharth:
lgtm

FelixGV:
Thanks, I merged it in.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -196,12 +196,17 @@ private Configuration getConfiguration(String sourceFileUrl, String hadoopConfig
 
             String security = config.get(CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION);
 
-            if(security == null || !security.equals("kerberos")) {
-                logger.error("Security isn't turned on in the conf: "
-                             + CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION + " = "
-                             + config.get(CommonConfigurationKeys.HADOOP_SECURITY_AUTHENTICATION));
-                logger.error("Please make sure that the Hadoop config directory path is valid.");
-                throw new VoldemortException("Error in getting Hadoop filesystem. Invalid Hadoop config directory path.");
+            if (security == null) {
```
arunthirupathi: Nitpick : Can we make this a table based method and only warning for the simple ? This code kind of looks little awkward with check for simple and throws if not kerberos, which does the right thing, but little harder to read.

A code like valid options are simple and kerberos. If not throw.
If option equals simple, log a warning is easier to understand.

FelixGV: Sure, let me re-jiggle that around a bit.

FelixGV: Here's a new version of that block of code, which is a bit simpler: 10b2c76b55de9c38b8ace77f792784b4e0641fa7

LMK if that's okay for you.

FelixGV: Easier to look at it this way: https://github.com/voldemort/voldemort/blob/10b2c76b55de9c38b8ace77f792784b4e0641fa7/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java#L199

arunthirupathi: Looks much better than what I suggested. Thanks for making this change.

FelixGV: Ok thanks, I merged it in.

PR #300, opened 2015-08-14
Removed MemLock class and related configs.

FelixGV:
It turns out our mlock call always failed and we've been running fine without it anyway, so there's no point in keeping that code around.


singhsiddharth:
lgtm

PR #302, opened 2015-08-20
Add a 'binary' format to the fetch-entries admin command so that it w…

jwlent55:
…ill create

a file compatible with the update-entries command.  For consistency sake add
this same format to the fetch-keys command.

My testing indicates that the fetch-entries command no longer produces a file format consistent with what the update-entries command expects.  To address this I defined a new output format called 'binary' that does match what the update-entries command expects (the old 'non ascii' format).  I considered several options:

1) Modify 'hex' format back to what it was before commit 2e880fcfc8 (I think) redefined the 'non ascii' format (which became the 'hex' format).
2) Modify the update-entries command to read the current 'hex' format.
3) Add a new 'binary' format consistent with the old 'non ascii' format.

I selected the third option because it preserved all the existing behavior.

Adding the 'binary' format (back) to the fetch-keys command was a somewhat arbitrary choice.

### DIFF
--- a/src/java/voldemort/tools/admin/AdminParserUtils.java
+++ b/src/java/voldemort/tools/admin/AdminParserUtils.java
@@ -76,6 +76,8 @@ public class AdminParserUtils {
     // defined argument strings
     public static final String ARG_FORMAT_HEX = "hex";
     public static final String ARG_FORMAT_JSON = "json";
+    // This format creates a file compatible with the update-entries command
+    public static final String ARG_FORMAT_BINARY = "binary";
 
     /**
      * Adds OPT_ALL_NODES option to OptionParser, without argument.
@@ -170,9 +172,9 @@ public class AdminParserUtils {
      * @param required Tells if this option is required or optional
      */
     public static void acceptsFormat(OptionParser parser) {
-        parser.accepts(OPT_FORMAT, "format of key or entry, could be hex or json")
+        parser.accepts(OPT_FORMAT, "format of key or entry, could be hex, json or binary")
               .withRequiredArg()
-              .describedAs("hex | json")
+              .describedAs("hex | json | binary")
               .ofType(String.class);
     }
 
--- a/src/java/voldemort/tools/admin/command/AdminCommandStream.java
+++ b/src/java/voldemort/tools/admin/command/AdminCommandStream.java
@@ -183,7 +183,7 @@ public class AdminCommandStream extends AbstractAdminCommand {
             stream.println("  stream fetch-entries -n <node-id>");
             stream.println("                       (-p <partition-id-list> | --all-partitions | --orphaned)");
             stream.println("                       (-s <store-name-list> | --all-stores) -u <url>");
-            stream.println("                       [-d <output-dir>] [--format json | hex]");
+            stream.println("                       [-d <output-dir>] [--format json | hex | binary]");
             stream.println();
             getParser().printHelpOn(stream);
             stream.println();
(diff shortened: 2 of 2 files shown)


arunthirupathi:
Thanks for the changes, this sounds fine to me. But if you are looking at using this for any real world application, you are dumping one large file without blocks or partitions and one byte corruption could make the entire file useless.

if this is for testing purpose this is fine.

Otherwise a better approach would be to add binary format option with a directory where a file can be created per partition/block with required data.

I will merge this in if it is for testing purpose only.

arunthirupathi:
I am going to pass on this review, as I feel the current option is misleading and is more detrimental to have this than live without one.

Unless other committers feel strongly otherwise, I intend to close this pull request.

FelixGV:
I'm curious to let @jwlent55 answer your question about the intended use case.

If it is really intended for testing purposes, then we might want to merge it in but simply print a warning stating that this functionality is intended for testing purposes, not production use.

-F

jwlent55:
We use it in get a snapshot of the data that can then be loaded into a dev environment.

One question. If you intend to reject this change, is your long term plan to remove the update-entries command?  Without an easy way to generate the format that the update-entries command expects what use is that command?

Without a way to generate a file that this command expects I don't see what use it provides.

arunthirupathi:
Interesting, did you try clusterForkLift tool ? It can help you to migrate
data from one cluster to other. But it requires a host which can access
both the source and the destination voldemort admin ports.

https://github.com/voldemort/voldemort/blob/master/src/java/voldemort/utils/ClusterForkLiftTool.java

We are not planning to change update-entries . I was just worried about the
fragility of the binary format.

Thanks,
Arun.

jwlent55:
Thanks for the tip.  I will look into the tool.  An initial search of the Wiki documentation did not turn up much, but, the Javadocs looks good.  It may, however, not meet all our needs (the direct source to destination copy may be problematic).  

We already build our own version of Voldemort (to include a couple of changes I do not think you would be interested in) so including this change too is no big deal.  I just thought the incapability between fetch and update was an oversight and a "fix" would be of general interest.  Since it has been that way for over a year now perhaps no one else expects them to be compatible.

FelixGV:
Hi @jwlent55,

The Forklift tool is something we use pretty regularly, so it definitely "works", though we don't use it for production use cases and it definitely has a few kinks. It is relevant to the discussion in that it is similar to your use case though it is not actually a proper snapshotting tool since it only transfers between live servers, it does not create an "at rest copy".

We never thought of Voldemort as having any snapshotting capabilities at all, though when you think about it, the fetch-entries API can indeed be used as such, like you intended...

In the past several months, we've been trying to cull features that are incomplete or buggy (either because they were never finished in the first place or because they have not been maintained properly as other changes happened over the years). We want to be careful about PRs which more readily expose features we do not consider production-ready, but at the same time, it would be ideal if the community did not need to maintain their own forks.

Let us think about it and come back to you...

Also, just out of curiosity, is your Voldemort fork publicly visible on GitHub, or is it internal?

-F

jwlent55:
Our fork of Voldemort is not even stored in git much less public on GitHub.  I have to patch over any changes to my own fork of Voldemort to push them to GitHub.  Creating a GitHub fork of Voldemort has been discussed, but, no traction so far.

FelixGV:
Hey @jwlent55,

I spoke with @arunthirupathi and we think it's fine to merge in the PR, so I did.

I did add some warnings to all stream admin commands to make it as explicit as possible that they're not intended for production use. If that changes in the future, then we can always remove the warnings at that time.

Please let us know if you have other changes you'd like to offload from your internal fork (:

-F

PR #303, opened 2015-08-22
Usability to shell and fork lift

arunthirupathi:
1) Fix the shell to display the utf8 and avro bytes in more human readable way.
2) Forklift tool fails silently or does the wrong thing . Made the forklift to abort and fail with error message instead.


FelixGV:
LGTM. I merged it in. Thanks @arunthirupathi !

PR #304, opened 2015-08-31
Improve unit test reliability

FelixGV:


FelixGV:
Thanks for the review @arunthirupathi. I have squashed and merged to master.

The ```BindException``` failures should be pretty much completely gone now. I think we still have a handful of flaky tests but they fail only rarely (compared to bind failures which almost certainly happened on every run)...

Yay!

PR #308, opened 2015-09-14
Fix the RocksDB iteration logic for keys and entries.  The problems a…

jwlent55:
…ddressed include:

1) Key and entry iterators failed to strip prefix from keys (if required).
2) RocksdbStorageEngineTest failed to test BdbStorageEngine (all tests).
3) Both Key iterators would miss first entry and then iterate off the end.
With these changes I was able to remove the Ignore annotation from 2 tests.

### DIFF
--- a/test/unit/voldemort/store/rocksdb/RocksdbStorageEngineTest.java
+++ b/test/unit/voldemort/store/rocksdb/RocksdbStorageEngineTest.java
@@ -55,9 +55,7 @@ public class RocksdbStorageEngineTest extends AbstractStorageEngineTest {
         props.setProperty("voldemort.home", "tmp/voldemort");
 
         voldemortConfig = new VoldemortConfig(props);
-        if(this.prefixPartitionId) {
-            voldemortConfig.setRocksdbPrefixKeysWithPartitionId(true);
-        }
+        voldemortConfig.setRocksdbPrefixKeysWithPartitionId(this.prefixPartitionId);
         this.rocksDbConfig = new RocksDbStorageConfiguration(voldemortConfig);
         this.rocksDbStore = (RocksDbStorageEngine) rocksDbConfig.getStore(TestUtils.makeStoreDefinition("test"),
                                                                           TestUtils.makeSingleNodeRoutingStrategy());
@@ -84,12 +82,4 @@ public class RocksdbStorageEngineTest extends AbstractStorageEngineTest {
     @Ignore
     @Override
     public void testTruncate() throws Exception {}
-
-    @Ignore
-    @Override
-    public void testKeyIterationWithSerialization() {}
-
-    @Ignore
-    @Override
-    public void testIterationWithSerialization() {}
 }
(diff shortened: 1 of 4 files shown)


jwlent55:
Just starting to test with the RocksDB storage engine and first want to get the unit tests passing.  I don't think the single partition iteration logic is covered by the existing RocksDB unit tests, but, I did test that manually.

FelixGV:
That's awesome @jwlent55!

I'll probably be able to review and merge this in later today.

Please keep us posted, here or on the mailing list, about your progress with RocksDB!

We're very interested to hear about how that goes (:

-F

test/unit/voldemort/store/rocksdb/RocksdbStorageEngineTest.java
```diff
@@ -85,11 +83,9 @@ public void tearDown() throws Exception {
     @Override
     public void testTruncate() throws Exception {}
 
-    @Ignore
```
FelixGV: I think here you need to completely remove the override, otherwise, you are still overriding that particular test with an empty implementation.

FelixGV:
Hi @jwlent55,

I don't understand what you mean by your second point:

> 2) RocksdbStorageEngineTest failed to test BdbStorageEngine (all tests).

Can you elaborate?

-F

jwlent55:
Ooops.  Regarding the comment:

2) RocksdbStorageEngineTest failed to test BdbStorageEngine (all tests).

Cut'n'paste error (and vague).  Let me try again with more details

2) RocksdbStorageEngineTest failed to test RocksDbStorageEngine (all tests).  Instead both times the test suite runs I think it tests the subclass PartitionPrefixedRocksDbStorageEngine.

The current code defaults the rocksdbPrefixKeysWithPartitionId boolean to true:

VoldemortConfig.java
        this.rocksdbPrefixKeysWithPartitionId = props.getBoolean("rocksdb.prefix.keys.with.partitionid",
                                                                 true);

But in the unit test only overrides the value in the unit test suite if "true" was passed in:

RocksdbStorageEngineTest.java

if(this.prefixPartitionId) {
            voldemortConfig.setRocksdbPrefixKeysWithPartitionId(true);
        }

I just changed this to always set the boolean based on what is passed in.

FelixGV:
Ah ok, sounds good!

Would you mind creating a new (squashed) commit with the two tests you fixed completely removed (rather than leaving the empty ```@Override``` function) and a commit message which references RocksDB rather than BDB?

I'll merge it in when it's ready.

Thanks a lot!

-F

test/unit/voldemort/store/rocksdb/RocksdbStorageEngineTest.java
jwlent55: Agreed.  My mistake.  I did my testing using a version of the unit test code that did not include those ignore/overrides.  I only saw that code when I patched it over to git.  In a brain dead move I just removed the Ignore.

jwlent55:
Will do and I may also roll in a small change to the RocksdbStorageEngineAPITest for the rocksdbPrefixKeysWithPartitionId boolean setting issue.

jwlent55:
I have updated the pull request commit to address your code and commit text comments.  Thanks for the review. Let me know if you want more changes and/or I did not create the new squashed commit correctly.

FelixGV:
Thanks a lot @jwlent55. LGTM (:

FelixGV:
This is merged in. Thanks @jwlent55 (:

PR #309, opened 2015-09-22
add additional keepalive settings to avoid client connection leaks

jalkjaer:
Extended server.properties to include booleans 
* nio.connector.keepalive 
* nio.admin.connector.keepalive

This will set SO_KEEPALIVE on the client connections in order to have the OS automatically close half open client connections


arunthirupathi:
Thanks for the contribution, will merge this in. Sorry for taking longer time than expected.

PR #313, opened 2015-09-25
Proper interruption of BnP hooks.

FelixGV:
In the previous implementation, jobs would wait until the next iteration of the heartbeat hook loop (up to 60 seconds) before terminating.


FelixGV:
Thanks for the review @arunthirupathi, I merged it in.

PR #317, opened 2015-10-01
Add heart beat for NIO Selector

arunthirupathi:
On Voldemort Server, when there is a disk crash the selector
gets hung. Acceptor still keeps on accepting sockets and
after a while the file descriptor limit is reached
and everything is in a hung state. No JVM stack or heap dump.

Now a selector heart beat is added and if the selector is past the
max heart beat time ( defaults to 3 minutes) the acceptor will
stop assigning connections to this selector. If all selectors
are unhealthy ( past max heart beat time) acceptor will close
the socket so that client can recover from these errors faster.

### DIFF
--- a/test/common/voldemort/ServerTestUtils.java
+++ b/test/common/voldemort/ServerTestUtils.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.http.client.HttpClient;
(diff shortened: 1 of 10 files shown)


test/common/voldemort/ServerTestUtils.java
```diff
@@ -172,6 +179,22 @@ public static AbstractSocketService getSocketService(boolean useNio,
...
+                                                         int coreConnections,
+                                                         int maxConnections,
+                                                         int bufferSize) {
+        return getSocketService(useNio,
+                                requestHandlerFactory,
+                                port,
+                                coreConnections,
+                                maxConnections,
+                                bufferSize,
+                                TimeUnit.MILLISECONDS.convert(3, TimeUnit.MINUTES));
```
FelixGV: I think you already have this default value set in ```VoldemortConfig```, right? You're hard-coding it in two other places. Can we use ```VoldemortConfig.getNioSelectorMaxHeartBeatTimeMs()```?

test/unit/voldemort/server/socket/NioSelectorManagerTest.java
```diff
@@ -0,0 +1,160 @@
...
+    private int port;
+    private ClientRequestExecutorPool pool;
+    private AbstractSocketService socketService;
+    private RequestHandlerFactory factory;
+    private SocketDestination dest1;
+    private SleepyStore<ByteArray, byte[], byte[]> sleepyStore;
+
+    private final static String STORE_NAME = "test";
+
+    // This can be never below the selecto poll time of 500 ms.
```
FelixGV: *selector

FelixGV:
LGTM. Left minor comments.

PR #318, opened 2015-10-02
Disk quota child branch

bhasudha:
- Moving quota setting logic from client to server
- Setting a default disk quota that can be set via server configs. If no quota is set, -1 is assumed for default disk quota, which will mean no disk quota checks will be done. This ensures that, introducing the disk quota check doesnt forcefully make users to use this feature.
- fix tests


FelixGV:
LGTM. Made a few suggestions. Please consider squashing all changes into a single commit (or two, if you prefer keeping the delete quota fix separate), with clean commit message(s) so our git history is easy to reason about.

Thanks a lot for hunting down these bugs!

FelixGV:
Thanks @bhasudha, feel free to merge in at your earliest convenience.

PR #319, opened 2015-10-05
Hadoop 2.6.0

dartiga:


FelixGV:
Thanks for the PR @dartiga,

I'm not sure we'll be able to merge this in in the near future. Unfortunately, we are stuck on Avro 1.4 here at LinkedIn and upgrading to 1.7 is non-trivial because of the incompatible API changes.

Ideally, we should have some kind of shim layer, where Voldemort can work with either Avro 1.4 or 1.7, so that the open-source community can move forward instead of getting pulled back by Voldemort's ancient dependencies... Which Avro version to use could be detected automatically at run-time and/or it could be decided at build time by bundling the correct module.

As for PB and Hadoop, it may be not as big a deal to upgrade them, but we would still probably want to do some testing just to make sure there are no regressions (in terms of performance or otherwise).

On the upside, this change is pretty minimal so it should apply fairly cleanly on any version of Voldemort, so if you want to maintain your branch with upgraded dependencies, it should represent a minimal amount of effort. We could even pull the upgraded branch into the official repo periodically so it is available there as well.

My apologies that this is not a smooth PR for us...

-F

FelixGV:
If we do move to a model where Voldemort supports multiple dependency versions, with the appropriate code bindings hooked up in a flexible way, then we should probably consider removing the auto-generated PB code completely, and have it be generated during every build. This way, the correct version would be generated, as per build configs. PB code-gen is super quick anyway, so there is no downside in terms of build time...

-F

dartiga:
It's fine, np! I'll keep my branch :)

Thank you!

PR #320, opened 2015-10-16
Implement the truncate function in the RocksDB storage engine. This f…

jwlent55:
…unctionality is required by

the delete store command.  It is implemented using a non default Column Family named after the store.
This approach could be extended to allow multiple stores to be supported by a single RocksDB
database. It is, however, not backwards compatible with the existing code. A data migration
would be required to support existing stores.

### DIFF
--- a/src/java/voldemort/store/rocksdb/PartitionPrefixedRocksDbStorageEngine.java
+++ b/src/java/voldemort/store/rocksdb/PartitionPrefixedRocksDbStorageEngine.java
@@ -4,6 +4,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.NoSuchElementException;
 
+import org.rocksdb.ColumnFamilyHandle;
+import org.rocksdb.ColumnFamilyOptions;
 import org.rocksdb.RocksDB;
 import org.rocksdb.RocksIterator;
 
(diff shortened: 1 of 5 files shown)


jwlent55:
The use of Column Families to implement this functionality may be controversial. Fire away, I have my shields up.

FelixGV:
Hi @jwlent55,

Sorry for the late reply...

Since we never officially announced support for RocksDB yet, I think we are probably still ok with backwards-incompatible changes, if they are well thought out.

Can you share your your assessments and your thoughts on why you think this is the right implementation strategy? Do you foresee any tradeoffs?

-F

jwlent55:
My first attempt was using the default column family, but, RocksDB would not let me drop the default column family.  My only other thought was implementing truncate using a big hammer like deleting DB, manually deleting the files, and them recreating the DB. There may me a better way.

jwlent55:
Just amended this commit to add a fix for issue that was uncovered when testing with a Mac build.  A call to storeHandle.dispose() was added to the truncate and close methods.  The Mac build runs with  assets enabled and the code was not happy that close was called while there was still a live reference to the column family handle.

jwlent55:
Had to amend this commit one more time.  This is a continuation to my previous amend.  That one addressed the typical case where there is at least a small time gap between the open and close of a store.  However, the Unit Tests often delete the store very soon after they create it.  Specifically before the JVM can clean up the handle to the default column family handle (i.e. before the GC can realize there is no live reference to it and its finalizer runs). This amend disposes of it immediately (since I am not using it for anything).

Note that this problem is only observable with asserts enabled in the RocksDB code (DEBUG_LEVEL >= 1).  Not sure why the Mac library is built that way by default, but, it helped uncover this issue.  I also don't think this is a "real" issue (i.e. there was no memory leak).  Still my original code depended (without really knowing it) on the finalizer to release some C++ resources.  Not good form.

FelixGV:
LGTM. I'll merge it in later today unless you have anything else you wish to sneak in.

Thanks a lot @jwlent55 (: !

-F

FelixGV:
I merged this in.

Thanks again for your contribution (: !

-F

jwlent55:
You are more than welcome.  We are glad there is another storage engine option to choose from.  One thing I have not pushed back is our default configuration values which we are currently setting via the server.properties file.  FYI...
```
# RocksDB 
rocksdb.prefix.keys.with.partitionid=false

# RocksDB Database Configuration
rocksdb.db.options.max_background_compactions=1

# RocksDB Column Family Configuration
rocksdb.cf.options.compression=kSnappyCompression
rocksdb.cf.options.max_write_buffer_number=3
rocksdb.cf.options.compaction_style=kCompactionStyleUniversal
rocksdb.cf.options.write_buffer_size=16777216
rocksdb.cf.options.level0_file_num_compaction_trigger=8
rocksdb.cf.options.level0_slowdown_writes_trigger=16
rocksdb.cf.options.level0_stop_writes_trigger=24
```
The key ones from a performance point of view in our configuration seemed to be:
* compaction_style=kCompactionStyleUniversal
* write_buffer_size=16777216
I would like to investigate the options more, but, ...

PR #321, opened 2015-10-16
Protection against data corruption.

FelixGV:
If index or data files are somehow corrupted, or mismatched (meaning
that the index file from one store/version gets used in combination
with a data file for another store/version), then the server gets into
bad buffer allocation problems, causing IllegalArgumentExceptions and
potentially OOMing in the process. This commit makes it easier to
debug the issues and protect against the OOMs. Changes include:

- Disallow allocation of negative size buffers.
- Disallow allocation of excessively large value buffers (configurable,
  max 25 MB by default).
- Disallow allocation of a buffer which would go past the data file's
  length.
- Server-side logs now print out the file name and key (in hex) which
  triggered the problem.
- Server-side logs now print out the socket name and stacktrace when
  SelectorManagerWorker catches a Throwable.
- Client-side will report an "internal server error" when any of the
  bad allocation scenarios described above happen.
- Slight optimization in the allocation strategy while reading the
  key/values in a data file.

### DIFF
--- a/src/java/voldemort/store/readonly/chunk/ChunkedFileSet.java
+++ b/src/java/voldemort/store/readonly/chunk/ChunkedFileSet.java
@@ -42,6 +42,7 @@ public class ChunkedFileSet {
 
     private int numChunks;
     private final int nodeId;
+    private final int maxValueBufferAllocationSize;
     private File baseDir;
     private final List<Integer> indexFileSizes;
     private final List<Integer> dataFileSizes;
@@ -58,7 +59,8 @@ public class ChunkedFileSet {
 
     public ChunkedFileSet(File directory,
                           RoutingStrategy routingStrategy,
-                          int nodeId) throws IOException {
+                          int nodeId,
+                          int maxValueBufferAllocationSize) throws IOException {
 
         this.baseDir = directory;
         if(!Utils.isReadableDir(directory)) {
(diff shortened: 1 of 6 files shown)


FelixGV:
Hi @gnb, @arunthirupathi,

I addressed the comments you made in e7db8278fb8a452ce06916e268d240b54eeb11ed and updated the commit message accordingly.

Please review the latest code.

-F

src/java/voldemort/store/readonly/chunk/ChunkedFileSet.java
```diff
@@ -544,47 +552,69 @@ public int getChunkForKey(byte[] key) {
...
-                    int keySize = sizeBuffer.getInt(ByteUtils.SIZE_OF_SHORT);
+                    int keySize = headerBuffer.getInt(ByteUtils.SIZE_OF_SHORT);
 
                     // Read the value size
-                    int valueSize = sizeBuffer.getInt(ByteUtils.SIZE_OF_SHORT
-                                                      + ByteUtils.SIZE_OF_INT);
+                    int valueSize = headerBuffer.getInt(ByteUtils.SIZE_OF_SHORT + ByteUtils.SIZE_OF_INT);
 
                     do {
+                        if (valueLocation >= (fileSize - headerSize)) {
```
FelixGV: @gnb, I added the check you requested for not going past the data file's length here.

FelixGV:
FYI, your new test passes but this change introduces a regression in ```HadoopStoreBuilderTest.testHadoopBuild()```. I will investigate that before merging it in.

FelixGV:
I fixed the bug in this amended commit on another branch: https://github.com/FelixGV/voldemort/commit/b5e969c18d355fd0773b321d2465cedb2d55dc61

I'm re-running the tests one more time and planning to merge this in afterwards.

-F

FelixGV:
This is merged in.

PR #322, opened 2015-10-22
Removed the set quota admin op from BnP.

FelixGV:
The default quota for new stores is now controlled by a server-side config, so this is redundant (and harmful to users who do not wish to use quotas in this manner).

@bhasudha, can you review this, since you know most about disk quotas? I thought this is the way we intended the quotas to work, but just in case there might be side-effects I'm not thinking of... Thanks (:


FelixGV:
Thanks Arun, will do (:

FelixGV:
Merged in. Thanks for the review @arunthirupathi (:

PR #323, opened 2015-10-22
This change avoids needlessly recreating symlinks.

FelixGV:
There are cases (i.e.: emergency situations) where we may want to chmod -w a data directory and recreating the symlinks systematically then prevents the server from being able to start up. This change inspects the symlink which is already present and avoids recreating it if it's already pointing at the right destination.


arunthirupathi:
Looks good to me.

FelixGV:
Squashed and merged https://github.com/FelixGV/voldemort/commit/49beabbc94755af04777a1d2e8ba2c3f7d31a6c9.

PR #324, opened 2015-10-24
Client do not remove dead connections

arunthirupathi:
Added more tests, this is ready to be merged in.

When Server grace fully closes a connection, client ignores the
connection close. After this an operation that gets this connection
fails with EOFException as read/write returns -1 which causes the
operation to fail. Mostly this is fine as the backup operation generaly
succeeds, but when a cluster is bounced due to upgrade or other reasons,
and if a client ends up caching all dead connections, it causes the
operation to fail.

After a client operation is done, previously the socket is registered
with the selector for no-intention. For the next operation we start with
write.

With this change, the socket will be left in read mode. So FIN packet
from the server will be processed and the socket will be closed. Any
connection checkout from the pool needs to validate the connection.
There is a race condition as the returned connection from the cache
could still be killed later, that is handled as well.

AysncRequestHandler close is called so the stream resources are freed.

Test is added to verify this situation.

### DIFF
--- a/src/java/voldemort/common/nio/AbstractSelectorManager.java
+++ b/src/java/voldemort/common/nio/AbstractSelectorManager.java
@@ -16,6 +16,7 @@
 
 package voldemort.common.nio;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.nio.channels.ClosedSelectorException;
 import java.nio.channels.SelectionKey;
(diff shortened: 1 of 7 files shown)


FelixGV:
LGTM. Thanks for the comments. Are you going to squash before committing?

PR #325, opened 2015-10-26
Output Checksum to Stats File

arunthirupathi:
1) Output Checksum to stats file, can be used for individual file
validation
2) Open the stats file in append only mode, so when a version is pushed
deleted and re-pushed we have some history in the stats file.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
@@ -1,12 +1,5 @@
 package voldemort.store.readonly.fetcher;
 
-import org.apache.commons.io.IOUtils;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.log4j.Logger;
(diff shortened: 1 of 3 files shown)


FelixGV:
Good idea @arunthirupathi (:

I wonder, since the stats file format fluctuates a fair bit, should we version it?

We could have another line before the header which says something like "Stats file format: v3" or something like that.

As far as I can tell, the format already changed twice, and this will be the third. In order to make our scripts more resilient to change, this could be a useful convention to introduce. Of course, we should also document what the format version to header mapping is, which I think would be good enough to simply do in the stats' file JavaDoc.

Alternatively, we could achieve the same by giving an extension to the stat files' names, i.e.: ```.../data/read-only/<store-name>/.stats/version-x.stats-v3```.

LMK what you think of embedding some format versioning info in the files' content or name.

-F

arunthirupathi:
@FelixGV  updated the commit , now the stats file will have a suffix -v<Version_Number>

FelixGV:
Thanks for adding the version format. I left two minor comments. Feel free to address them or not before merging in. Otherwise LGTM.

arunthirupathi:
Fixed.

PR #326, opened 2015-10-27
Fixed regression introduced in d25d169862cc12e599bb1c916d684d89b5b4bfd4.

FelixGV:
The Read-Only servers used to percolate exception messages back to Build and Push jobs, for easy debugging of Hadoop-related exceptions (i.e.: AccessControlException). As part of the commit above, the server begun swallowing the underlying exceptions' message, which makes debugging BnP jobs harder. This commit fixes the regression.

Also included some minor clean ups, such as JavaDoc improvements and @Deprecated annotations for the test-only functions.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -104,8 +104,11 @@ public class HdfsFetcher implements FileFetcher {
      * The keytab path is set to empty string, which triggers a different code path
      * in {@link HadoopUtils#getHadoopFileSystem(voldemort.server.VoldemortConfig, String)}.
      *
-     * TODO: Change visibility or otherwise ensure that only test code can use this...
+     * FIXME: Change visibility or otherwise ensure that only test code can use this...
+     *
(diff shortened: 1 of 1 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -299,18 +312,15 @@ private File fetchFromSource(String sourceFileUrl,
                 logger.error("Source " + path.toString() + " should be a directory");
                 return null;
             }
-        } catch(Exception e) {
+        } catch (Exception e) {
             if(stats != null) {
                 stats.reportError("File fetcher failed for destination " + destinationFile, e);
             }
-            String errorMessage = "Error thrown while trying to get data from Hadoop filesystem : ";
-            logger.error(errorMessage, e);
```
arunthirupathi: Can the error be still logged ? May be without the call stack or is it logged some where else ?

FelixGV: This is already logged higher up the stack in ```AdminServiceRequestHandler```, which results in duplicate logs.

FelixGV:
This is merged in.

PR #327, opened 2015-10-27
CopyStats Regression

arunthirupathi:
MessageDigest.digest() is read once, after it is read
the contents of the digest are reset.

Without knowing this, now started outputting checksum to stats file
which broke the normal flow.

Passing Around MessageDigest is dangerous because of this, so modified
the method to return byte arrays which are then passed around.


FelixGV:
Thanks for the fix! LGTM.

FelixGV:
I merged this one in.

PR #328, opened 2015-10-29
Added a validator for Read-Only files.

FelixGV:
Given a data directory, the ```ReadOnlyFileValidator``` class will check a few things:
- Index files are well formed:
  - Well sorted
  - no negative offset
- Data files are well formed:
  - Number of records for a given key hash is > 0
  - Key and value sizes are > 0
- Index and data files match together:
  - No gaps in data files, when traversing them with the offsets from index files
  - Raw keys in data files match the hashes found in their index files

It then prints a summary of the amount and kinds of problems found.

I am posting this here in case anyone is curious, but I might modify it further before being ready to merge it in, so please don't merge it in yet yourselves. Thanks.

### DIFF
--- a/src/java/voldemort/store/readonly/ReadOnlyFileValidator.java
+++ b/src/java/voldemort/store/readonly/ReadOnlyFileValidator.java
@@ -1,0 +1,405 @@
+package voldemort.store.readonly;
+
+import voldemort.utils.ByteUtils;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.util.Arrays;
+
+/**
+ * This class is used to iterate over index and data files to validate their integrity.
+ *
+ * N.B.: This tool only works with the Read-Only V2 format.
+ *
+ * May you never need it.
+ */
+public class ReadOnlyFileValidator {
+    public static void main(String[] args) {
+        if (args.length < 1) {
+            printInfo("Please include the following parameters: data_path [-v]");
+            printInfo("");
+            printInfo("data_path: Data directory for a store/version, or individual index or data file.");
+            printInfo("-v:        Activates the verbose mode (optional, default: off).");
+            printInfo("");
+            printDisclaimers();
+            System.exit(1);
+        }
+
+        String pathString = args[0];
+        boolean verbose = (args.length > 1 && args[1].equals("-v"));
+
+        File path = new File(pathString);
+
+        ReadOnlyFileValidator validator = new ReadOnlyFileValidator();
+
+        try {
+            if (path.isDirectory()) {
+                printInfo("Running in directory mode.");
+                validator.validateDirectory(path, verbose);
+            } else if (path.getName().endsWith(ReadOnlyUtils.INDEX_FILE_EXTENSION)) {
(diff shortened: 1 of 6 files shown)


arunthirupathi:
Do you think a check to read .metadata file and make sure the checksum matches between the data and index file is worth adding ?

That will help in identifying the disk corruption.

The way I see it is , these checks if errors out we have errors. But absence of errors does not mean they are good.

src/java/voldemort/store/readonly/ReadOnlyFileValidator.java
```diff
@@ -0,0 +1,363 @@
...
+        for (File file: files) {
+            if (file.length() == 0) {
+                if (verbose) {
+                    System.out.println("Skipping empty file: " + file.getName());
+                }
+            } else if (file.getName().endsWith(ReadOnlyUtils.INDEX_FILE_EXTENSION)) {
+                problemsDetectedInIndexFiles += validateIndexFile(file, verbose);
+                problemsDetectedInIndexAndDataFilesMatch += validateIndexAndDataFileMatch(file, verbose);
+            } else if (file.getName().endsWith(ReadOnlyUtils.DATA_FILE_EXTENSION)) {
+                problemsDetectedInDataFiles += validateDataFile(file, verbose);
```
arunthirupathi: Should you check for existence of .index file here ?

src/java/voldemort/store/readonly/ReadOnlyFileValidator.java
```diff
@@ -0,0 +1,363 @@
...
+                                                       String description,
+                                                       int keysRead) {
+        int cmp = ByteUtils.compare(previousByteArray, currentByteArray);
+
+        if (cmp < 0) {
+            // This is the expected case. The index file appears to be well-formed so far.
+            return 0;
+        } else if (cmp == 0) {
+            // The same key hash or offset appeared twice in a row. Unexpected!
+            System.err.println("We found two identical consecutive key " + description +
```
arunthirupathi: Should we print them only under verbose conditions ?

FelixGV: Yes, good point.

src/java/voldemort/store/readonly/ReadOnlyFileValidator.java
```diff
@@ -0,0 +1,363 @@
...
+            valueSize = ByteBuffer.wrap(currentValueSize).getInt();
+
+            if (verbose) {
+                System.out.println("Key #" + keysRead +
+                                   ", Key Size: " + keySize +
+                                   ", Value Size: " + valueSize);
+            }
+
+            if (keySize == 0) {
+                System.out.println("Key size 0!");
```
arunthirupathi: Same verbose here, also this uses System.out other uses System.err

FelixGV: Good catch, I'll refactor all of that.

FelixGV:
Thanks for the review @arunthirupathi. I amended my previous commit.

I added a bash script to easily run the ```ReadOnlyFileValidator```. I also added disclaimers about what this script does not do.

I'm good to go as far as merging this in.

arunthirupathi:
Looks good to me

PR #330, opened 2015-11-03
Forklift tool schema comparison

arunthirupathi:
1) ignore-schema-check was handled but not exposed as a parameter.
Seems like joptSimpleParser will error out, if the parameters are not
set up correctly.

2) Serializers are compared using string which resulted in lot more
error. Using the Schema comparison based on Avro and JSON now. Most of
the code here is auto generated (except for AVroGenericVersioned) where
I modified some code by hand.

3) Fixed invalid stores xml.

### DIFF
--- a/src/java/voldemort/serialization/SerializerDefinition.java
+++ b/src/java/voldemort/serialization/SerializerDefinition.java
@@ -129,8 +129,7 @@ public class SerializerDefinition {
         return compression;
     }
 
-    @Override
-    public boolean equals(Object obj) {
+    private boolean commonEquals(Object obj) {
         if(obj == this)
             return true;
         if(obj == null)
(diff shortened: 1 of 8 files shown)


src/java/voldemort/serialization/SerializerDefinition.java
```diff
@@ -139,11 +138,38 @@ public boolean equals(Object obj) {
...
+        boolean isEqual = commonEquals(obj);
+        if(!isEqual) {
+            return false;
+        }
+        SerializerDefinition s = (SerializerDefinition) obj;
+        return Objects.equal(this.schemaInfoByVersion, s.schemaInfoByVersion);
+    }
+
+    public boolean semanticEquals(Object obj) {
+        boolean isEqual = commonEquals(obj);
```
FelixGV: I don't understand why would you need ```commonEquals``` here?

What you're doing in the beginning of the ```semanticEquals``` is exactly the same thing that you are doing in the ```equals()``` function.

I think you can just leave the ```equals()``` as it was before, and depend on it from your new ```semanticEquals()``` function.

FelixGV: Oh no ok... I get it.

PR #331, opened 2015-11-03
Moved BnP's verifyOrAddStore() to the AdminClient.

FelixGV:
This makes it easier to leverage this more resilient/idempotent Add Store operation from other processes.


FelixGV:
FYI @arunthirupathi, I amended my commit after addressing your concern over the use of ```verifyOrAddStore``` with RW ```StoreDefinitions```. Also included some other minor refactorings.

FelixGV:
Merged. Thanks for the review @arunthirupathi.

PR #334, opened 2015-11-06
Refactoring lots of BnP code

FelixGV:
I am submitting this PR just to get a jump start on the review, but I'm not planning to merge it in as is (I will squash and write a better commit message before merging).

In summary, this is just the removal of dead and/or duplicate code. The individual commits have more details. I tested pushing Avro and JSON data and everything seems to work fine still.

This work is a stepping stone towards building partitions only once in BnP, which will reduce Hadoop resource utilization, as allow us to support pushing to heterogeneously-partitioned clusters.


FelixGV:
FYI: the squashed commit would be this one:

https://github.com/FelixGV/voldemort/commit/8e662fcc77a2492639810fa8e4516a4459623ffd

Feel free to review either one, whichever you think is easier to read.

-F

arunthirupathi:
Looks good to me. but worried about the regressions. Can you come up with a test plan and/or unit tests to reduce the chance of regressions ?

FelixGV:
Thanks for the review @arunthirupathi.

You're right that it is a bit of a hairy change, so it could potentially introduce regressions.

Unfortunately, it is a bit hard to do automated tests for MR jobs. There are ways of doing it (like MRUnit and others), but as far as I can tell we don't currently use any of that in the current Voldemort tests, and so the tests don't really take distribution into account.

We could unit test some of the components, such as the Partitioners with static input parameters coupled to expected output.

I am also already testing the job in real distributed mode with various parameters, so I could come up with a more strict test plan in that regard.

-F

FelixGV:
This PR is superseded by #356. Closing.

PR #335, opened 2015-11-06
Make VoldemortClientShell adminClient use user-defined properties file

stotch:
adminClient now uses the same ClientConfig as the SocketStoreClient so that it too can get the overrides passed in by a user defined config file.


arunthirupathi:
Looks good, please merge it in.

PR #338, opened 2015-11-11
Bug fix for BnP HA

FelixGV:
- StoreVersionManager.getDisabledMarkerFile() did not look for the right file name.
- Also added unit tests and better log messages for this.

Debuggability improvements in AdminServiceRequestHandler:
- More exceptions now logged with their full stacktrace on the server-side.
- The request info is now printed in a more readable format.

BnP job will now fail properly when clusters are inconsistent, rather than just exiting.


singhsiddharth:
lgtm

PR #340, opened 2015-11-12
add a customized workload that supports key size distribution and val…

jziub:
Hi,

I wanted to test Voldemort by replaying the workload trace I have but can't because the benchmark restricts the value size and key size as fixed numbered, and does not allow workloads loaded from external files. 

So I changed the benchmark test suite a little bit to allow variable key size and value size generated from a workload trace. Too, I extended the benchmark framework a little bit to allow customized workload provided by users. Details are as follows. I wonder if this feature can be integrated into the branch. Thanks. 

https://github.com/jziub/voldemort/commit/cf0c905e7fa1d4daf85bf75e4cb04cd660c928c3#diff-179c5aadb8e491e49f068c9cd438d084R442

### DIFF
--- a/test/integration/voldemort/performance/benchmark/Benchmark.java
+++ b/test/integration/voldemort/performance/benchmark/Benchmark.java
@@ -24,9 +24,10 @@ import java.util.List;
 import java.util.Vector;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.io.FileUtils;
+
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
-import org.apache.commons.io.FileUtils;
 import voldemort.ServerTestUtils;
 import voldemort.StaticStoreClientFactory;
 import voldemort.TestUtils;
@@ -121,13 +122,19 @@ public class Benchmark {
 
     public static final String LOCAL_SERVER_PROPERTIES = "local-server-properties";
 
+    public static final String WORKLOAD_TYPE = "workload-type";
+    public static final String KEY_VALUE_FILE = "key-value-file";
+    public static final String KEY_SEQUENCE_FILE = "key-seq-file";
+    public static final String DEFAULT_WORKLOAD_TYPE = "default-workload";
+    public static final String TRACE_WORKLOAD_TYPE = "trace-workload";
+
     private StoreClient<Object, Object> storeClient;
     private StoreClientFactory factory;
 
     private int numThreads, numConnectionsPerNode, numIterations, targetThroughput, recordCount,
             opsCount, statusIntervalSec;
     private double perThreadThroughputPerMs;
-    private Workload workLoad;
+    private BenchmarkWorkload workLoad;
     private String pluginName;
     private boolean storeInitialized = false;
     private boolean warmUpCompleted = false;
@@ -184,7 +191,7 @@ public class Benchmark {
         private VoldemortWrapper db;
         private boolean runBenchmark;
         private boolean isVerbose;
-        private Workload clientWorkLoad;
+        private BenchmarkWorkload clientWorkLoad;
         private int operationsCount;
         private double targetThroughputPerMs;
         private int opsDone;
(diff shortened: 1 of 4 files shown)


test/integration/voldemort/performance/benchmark/CustomizedWorkload.java
```diff
@@ -0,0 +1,243 @@
...
+import voldemort.performance.benchmark.generator.DiscreteGenerator;
+import voldemort.utils.Props;
+import voldemort.versioning.Versioned;
+
+/**
+ * This class implements a workload whose key size distribution, value size
+ * distribution and key access sequence are loaded from files. Thus it allows
+ * real world workload to be replayed.
+ */
+public class CustomizedWorkload implements IWorkload {
```
arunthirupathi: Customized is too generic of a name, can you please rename it to something else. TraceReplayWorkload or RecordedReplayWorkload or something else ?

test/integration/voldemort/performance/benchmark/Benchmark.java
```diff
@@ -642,6 +651,18 @@ public static void main(String args[]) throws IOException {
               .withRequiredArg()
               .describedAs(LOCAL_SERVER_PROPERTIES)
               .ofType(String.class);
+        parser.accepts(WORKLOAD_CLASS, "workload class (full name)")
+              .withRequiredArg()
+              .describedAs(WORKLOAD_CLASS)
+              .ofType(String.class);
```
arunthirupathi: Can you please add help message about what are the possible options of the Workload class ? I believe exposing the class is probably not convenient and sort of gives an impression that user can supply their workload. Not a big deal, but would be better if it can be renamed as WORKLOAD_TYPE and just let the type be former or the trace replay.

arunthirupathi:
Left some comments, thanks for the contribution. Will merge this in once the review comments are addressed.

FelixGV:
Very cool (:

I'm going to echo @arunthirupathi's comments. ```CustomizedWorkload``` sounds a bit too generic. Arun's suggested names look fine, or maybe just ```FileWorkload``` could be fine as well.

We don't really use the convention of prefixing interfaces with a capital I anywhere in the code base, so I think the new interface, ```IWorkload```, should be named differently as well. Maybe ```BenchmarkWorkload```, or just ```Workload```?

Thanks a lot for the contribution!

-F

test/integration/voldemort/performance/benchmark/Benchmark.java
jziub: So you don't want to allow a user to implement her own workload class? I thought this might be more flexible. For example, a user may simulate the time interval of requests sent based on traces.

I can certainly use workload-type and give it two options like default-workload and trace-workload.

arunthirupathi: I like your approach. But I believe using CLASS name in the option is misleading and sort of suggests reflection and other.

WORKLOAD_TYPE may be a better name in my opinion.

jziub: ok, does the following look good to you? I did use reflection to load the workload class anyway. 

parser.accepts(WORKLOAD_TYPE,
                       "class of the workload to use [optional, default is voldemort.performance.benchmark.DefaultWorkload]")

arunthirupathi: yeah.... that sometimes will confuse. Using reflection without providing a way to load the class path will confuse. 

I suggest giving WORKLOAD_TYPE default is default if given File, the other workload is used.

FelixGV:
Thanks for addressing the comments, @jziub (:

Can you please make sure your changes to the eclipse file is properly ignored / excluded from the commit?

Also, could you squash all your changes into a single commit?

I think this would be ready to be merged in afterwards.

-F

jziub:
@FelixGV sure, will do ASAP. thx.

jziub:
@FelixGV below is the new commit. let me know if you need other changes. thx.

https://github.com/jziub/voldemort/commit/84786ee3e96b60125799720190d127118be7fb56

FelixGV:
Thanks for the contribution @jziub ! Sorry for the long wait time, we got distracted by other pressing concerns. I have now merged in your contribution (:

PR #341, opened 2015-11-13
ReadOnlyReplicationHelperCLI: do not rely on down node

gnb:
When building a list of partitions to be copied between nodes
to restore a down node, don't expect to be able to fetch any
useful metadata from the down node.


arunthirupathi:
Looks good to me, although, I would rename the method as getCurrentStoreDefinitionsExcept instead of having the exceptNodeId in the parameter.

On the caller, it totally gives wrong meaning and you have put it in the comments for now.

gnb:
> I would rename the method as getCurrentStoreDefinitionsExcept

Fair call, I'll do that.

PR #342, opened 2015-11-13
ReadOnlyReplicationHelperCLI: do not rely on down node

gnb:
When building a list of partitions to be copied between nodes
to restore a down node, don't expect to be able to fetch any
useful metadata from the down node.

v2: better method naming per feedback from athirupathi


arunthirupathi:
looks good, please merge it in.

PR #344, opened 2015-11-16
Extend shell "preflist" command to show partitions.

gnb:


arunthirupathi:
Modifying StoreClient interface will create issues for existing implementers of this interface.
I believe StoreClient exposing getReponsibleNodes is already a bad leaky abstraction and the partition should not also be leaked in there.

I see a new pull request with more contained fix. So closing this one.

PR #346, opened 2015-11-17
Update metadata version

arunthirupathi:
Update the metadata version on store and cluster update so that the client bootstraps after the change.


FelixGV:
LGTM. The second commit seems pretty intimately tied to the first. Is it worth keeping them separate, or might it be more clear to squash them together?

arunthirupathi:
The second commit is dependent on the first.

The first one fixes the cluster version and the second one fixes the stores version. Will squash them into one and merge it.

PR #347, opened 2015-11-18
Introduced 'build.primary.replicas.only' mode in BnP.

FelixGV:
This is based on the refactorings in the previously submitted PR #334.

Intended only for early reviewing. Not ready to be merged in yet!

Before merging, I still want to:
1. Test a whole lot more.
2. Add a way for the BnP job to determine if it can run in 'build.primary.replicas.only' mode or not, based on server config.

-F


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -245,21 +255,67 @@ private File fetchFromSource(String sourceFileUrl,
...
+                    if (metadataFile.getDiskFileName().equals(ReadOnlyUtils.FULL_STORE_METADATA_FILE)) {
+                        // Then we are in build.primary.replica.only mode, and we need to determine which
+                        // partition sub-directories to download
+                        StoreDefinition storeDefinition = metadataStore.getStoreDef(storeName);
+                        Cluster cluster = metadataStore.getCluster();
+                        List<Integer> partitionsMasteredByCurrentNode = cluster.getNodeById(metadataStore.getNodeId()).getPartitionIds();
+                        RoutingStrategy routingStrategy = new RoutingStrategyFactory().updateRoutingStrategy(storeDefinition, cluster);
+                        for (int partitionId = 0; partitionId < cluster.getNumberOfPartitions(); partitionId++) {
+                            // For each partition in the cluster, determine if it needs to be served by the current node
+                            boolean partitionIdIsServedByCurrentNode = false;
```
arunthirupathi: I believe computing the partitionIds used by the node should be a method by itself, which just returns a list of integers partition id to be fetched by current node. 

This makes the code harder to reason about.

FelixGV: I agree. I think it belongs in the ```RoutingStrategy``` implementation, but I'm not exactly sure where would be the best place to put it.

FelixGV: I moved this out into its own function. Also moved more of the quota code into ```checkIfQuotaExceeded```.

FelixGV:
This PR is superseded by #356. Closing.

PR #348, opened 2015-11-19
For testing BnP with sWebHdfs, let Voldemort node can enable or disable SSL separately.

squarY:
In order to enable and disable SSL support in Voldemort node. Add 3 properties in configuration and replace source file url before fetching if needed.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -18,10 +18,6 @@ package voldemort.store.readonly.fetcher;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLConnection;
-import java.net.URLStreamHandler;
 import java.text.NumberFormat;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -34,7 +30,6 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.log4j.Logger;
 
-import sun.net.www.protocol.file.Handler;
 import voldemort.VoldemortException;
 import voldemort.client.ClientConfig;
 import voldemort.client.protocol.admin.AdminClient;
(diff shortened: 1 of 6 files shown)


FelixGV:
Hi @squarY,

I have a few comments:
1. Can we make this new code generic? i.e.: I would like to be able to replace any protocol suggested by the BnP job to any other arbitrary protocol (not just transform for webhdfs to swebhdfs). Check the code here for an implementation that already does that: https://github.com/voldemort/voldemort/blob/6050ab929a807aa32397ce2aee752320cde9d763/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortSwapJob.java#L97-L113
2. If we do it that way, then I think we don't need the ```readonly.hdfs.ssl``` config at all, right?
3. (No need to do that right now but) before you merge the code in, I'd like us to squash it into one commit with relevant/concise message.

Thanks!

-F

squarY:
@FelixGV 
1. I have updated codes to make this feature more generic. Please check them out.
2. I think keep a property as switcher will let code and config more clearly. So I keep "readonly.modify" as boolean property.
3. Will do after passing review.

Thanks
Yan

squarY:
@gnb 
Thanks for your advice. Replaced codes of parsing by java.net.URL.

FelixGV:
Hmm... Can we just assume that the protocol should not be modified if ```readonly.modify.protocol``` is null (i.e.: undefined in the config)? And same for the port (i.e.: modify it only if ```readonly.modify.protocol``` is not null)?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -210,6 +210,36 @@ public File fetch(String sourceFileUrl,
         }
     }
 
+    /**
+     * Replace swebhdfs protocol and port in sourceFileUrl when SSL is not enabled in this node.
+     */
+    protected String modifyURLbyConfig(String sourceFileUrl) {
```
FelixGV: Do you think it'd make sense to move this code to a Utils class, and take newProtocol and newPort as input parameter?

This way, we could call the same code from HdfsFetcher as well as from the ```VoldemortSwapJob``` (here: https://github.com/voldemort/voldemort/blob/6050ab929a807aa32397ce2aee752320cde9d763/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortSwapJob.java#L97-L113).

squarY: Will do. Can I move this feature to https://github.com/voldemort/voldemort/blob/master/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/VoldemortUtils.java

FelixGV: Yeah, I think that should be fine (:

squarY:
@FelixGV 
Done. Please check the change.

Thanks
Yan

FelixGV:
LGTM! Thanks for making this change (:

Feel free to squash and merge.

PR #351, opened 2015-11-25
Multi module gradle build for voldemort

tempredirect:
Each contrib directory now becomes a separate project (although the
no new build.gradle files have been created). This makes the import
into Intellij better as it can setup the contrib source roots
correctly.

Tasks remaining:

 - [x] check eclipse import
 - [x] verify that bnp jar is being generated correctly


tempredirect:
Hi I've just rebased this to work on the latest master. Any change somebody can have a look at the bnp jar, as I don't really know how to verify that I haven't broken it.

arunthirupathi:
Apologize for not taking a look at it. Here is what I intend to do and if
you can help me that will be great.

Important:
      Difference between jar and contrib.jar from old way of building and
new way of building. Checkout the old master, build the jars. Checkout your
patch build the new jars. unpack (jar xvf) them. Use meld or other
directory diffing tool to compare both of the jars. The differences, if
there are any needs to be understood or ignored.

IDEs. This is OK if the IDE integration breaks or does not work smoothly,
but it is still good to verify the Eclipse or IDEA or atleast one of them.
Current one works for Eclipse, IDEA not so much.

It would take 2-3 hours for me to do that and I haven't found time to do
it, as weekends are quiet busy these days for me.

Community if one of you can do this, that will be awesome as well.

Thanks,
Arun.

tempredirect:
I'll do the jar diff'ing tomorrow, will upload a gist of the results

tempredirect:
have compared the contents of the jars. 

The voldemort-1.10.14.jar are basically the same, while they don't have the same checksum, they are the same size and contain the same files and the Manifests are the same.

The contrib jar is a little different see:

https://gist.github.com/tempredirect/e7420c37c93b6d2f69fc60310f36eb71

there are two core differences:

 * The new one doesn't include the test classes from contrib. I think this is an improvement
 * The new one has the package.html files ... not 100% certain why contrib has them but the main jar doesn't... will investigate a bit further

arunthirupathi:
Awesome, this is very great stuff, thanks for testing it. When you are comfortable I will merge it in.

tempredirect:
Have just cleaned up some of the setup for the contrib projects. I'm happy with it.

tempredirect:
Have checked the bnp jar in the same way as the jar and contrib jars

https://gist.github.com/tempredirect/3e94bc090b3169ee0323d2f4033218ca

only difference is the absence of the test classes

arunthirupathi:
I tried to merge this, but after this change, none of the shell commands work. The lib directory needs to be populated with all the dependencies, as both the server and client command depends on the lib directory in the class path to run properly.

[athirupa@athirupa-ld1 voldemort]$ bin/voldemort-server.sh config/single_node_cluster/
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/log4j/Logger
	at voldemort.common.service.AbstractService.<clinit>(AbstractService.java:36)
Caused by: java.lang.ClassNotFoundException: org.apache.log4j.Logger
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 1 more

Also if possible, I would recommend you to squash the changes. If not, it is ok, I will squash and merge them.

Thanks for your contribution again.

build.gradle
```diff
@@ -300,63 +301,55 @@ war {
 
 assemble.dependsOn copyDeps
```
arunthirupathi: Do you want to leave this as jar.dependsOn. 

For example, we run ./gradlew clean tar, which produces the tar ball of voldemort and we send the tar ball to other machines for running the command line and testing. now build and tar will probably work, but that is a change in behavior.

arunthirupathi:
I made on more pass through of the commit. Thanks for taking your time to make this change. 

Here are the 2 asks from me
1) tar task now does not build the tarball. Can you make the tar task build a tar ball if possible
2) supress the javadoc error if possible.

arunthirupathi:
You can look at this and let me know if you are fine with this change for the 2 changes I suggested.

https://github.com/arunthirupathi/voldemort/commit/2e0e9cf58a5fd9dd386dcbed5f1af2e497290e2c

tempredirect:
@arunthirupathi sorry about that...have taken your commit and just added the ability to enable the javadoc generation via a property ... this will actually help me internally at work as we like to publish source and javadoc jars to are internal repo.

tempredirect:
after a small bit of hackery, (i think) I have managed to squash into a single commit without causing a total mess.

arunthirupathi:
Thanks @tempredirect , give us another week and I will run it through other collaborators and merge this in.

FelixGV:
I'm not a Gradle wizard, but it LGTM.

arunthirupathi:
I merged this in. But there are 3 outstanding issues.

gradle task junitAll now run 1,776 tests instead of 1,864 tests  before.

Previously, I can run single test by ./gradlew test
-Dtest.single=ZoneShrink* , now though this runs the test, it also reports
the task as failure as it repeats it as contrib. with the message.
* What went wrong:
Execution failed for task ':contrib-collections:test'.
> Could not find matching test for pattern: *SimpleSocketPool*

Eclipse integration though lists most files, reports some files as not part
of the project and hence not searchable. But reverting this to the old hack
brings in all the files. But this is a lower priority.

I want to fix the number of tests though, others are inconvenience, that we
want to fix eventually.

Thanks,
Arun.

tempredirect:
the second is easier

    ./gradlew :test -Dtest.single=ZoneShrink*

note the `:` prefixing the task this gives gradle the hint to run only the test task in the root project, rather than 'test' in all projects.

The junitAll task doesn't exist in the contrib projects...prob best to just to add one there. Shouldn't take me too long.

tempredirect:
just been having a look at the junitAll build ... and it does run the tests from the contrib projects (in separate runs.  Where are you reading of the number of tests from? 

I'm guessing it's a CI job somewhere.. if so can you tell me where it is looking for the junit output?

Will have a look at the eclipse thing.... do people really still use eclipse?

arunthirupathi:
Gareth,
        I am looking at the report output under the aggregatedUnit.
Probably it will be running it, but the report does not include it. Thanks
for taking a look, appreciate that.

Thanks,
Arun.

tempredirect:
that's a bit weird, mine does include the contrib tests, any chance you could zip up the reports directory and email it to me?  gareth@logicalpractice.com

tempredirect:
I'm struggling to find the issue with the ecplise stuff.. managed to generate and import the project using the `./gradlew ecplise` and also using the 'Buildship' gradle plugin both seem to work, can run tests etc

arunthirupathi:
You are right. The new build runs more tests 1883.

But the hudson boxes we use, looks for the XML generated and it does not
find the *.xml from the contribJunit which causes it to report lesser
tests. Though the actual tests in the aggregatedJunit is actually more. My
bad, sorry for overlooking this.

On high level the old build directory had
contribJunit-results
junitLong-results
reports
test-results

The new one had contribJunit-results missing, which causes it to report
less tests than it actually ran. I am working on down sizing the zip and
can send it across if you are still interested.

For the Eclipse issue, try going to the file in contrib, though it is
actually part of Eclipse, you can't see call hierarchy or references. Go to
the file HdfsFetcher and see the call hierarchy for the constructor. I get
a popup saying that "The resource is not part on the build Path of a Java
Project".

Thanks for helping to follow through the issues, appreciate that.

Thanks,
Arun.

tempredirect:
is it possible to change the CI job to look for **/test-results/*.xml if not I'm sure I can copy the *.xml in to a contribtest-results directory.

in eclipse do you have four additional projects in eclipse

arunthirupathi:
For CI, this is the pattern we have today build/*-results/*.xml, since this
is XML only, not a high priority.

For Eclipse, I did not notice additional projects and once I included them,
it is fine now. Thanks for fixing the eclipse.

Appreciate your quick response.

Thanks,
Arun.

arunthirupathi:
The CI issue was resolved by modifying the test-results XML pattern to.

build/*-results/*.xml,contrib/*/build/*-results/*.xml

Thanks,
Arun.

FelixGV:
Uh oh...

I think this build change broke the scripts under ```bin/```... The following does not work anymore:
```
[fgv-ld1:voldemort]$ ./gradlew clean jar
...
BUILD SUCCESSFUL
...
[fgv-ld1:voldemort]$ ll dist/
total 2.7M
drwxr-xr-x  4 fgv 4.0K May  2 09:38 ./
drwxr-xr-x 25 fgv  12K May  2 09:37 ../
drwxr-xr-x  3 fgv 4.0K May  2 09:38 classes/
drwxr-xr-x  2 fgv 4.0K May  2 09:38 resources/
-rw-r--r--  1 fgv 2.6M May  2 09:38 voldemort-1.10.15.jar
[fgv-ld1:voldemort]$ ./bin/voldemort-shell.sh 
Error: Could not find or load main class jline.ConsoleRunner
```

We used to get more subdirectories generated under ```dist/```, so that's probably why the scripts can't pick up the right binaries.
```
[fgv-ld1:voldemort]$ git checkout release-1.10.14-cutoff 
Previous HEAD position was c8c457f... Releasing Voldemort 1.10.15
HEAD is now at 6809fce... Releasing Voldemort 1.10.14
[fgv-ld1:voldemort]$ ./gradlew clean jar
...
BUILD SUCCESSFUL
...
[fgv-ld1:voldemort]$ ll dist/
total 5.4M
drwxr-xr-x  6 fgv 4.0K May  2 09:40 ./
drwxr-xr-x 26 fgv  12K May  2 09:40 ../
drwxr-xr-x  3 fgv 4.0K May  2 09:40 classes/
drwxr-xr-x  3 fgv 4.0K May  2 09:40 contrib-classes/
drwxr-xr-x  3 fgv 4.0K May  2 09:40 resources/
drwxr-xr-x  5 fgv 4.0K May  2 09:40 testclasses/
-rw-r--r--  1 fgv 2.6M May  2 09:40 voldemort-1.10.14.jar
-rw-r--r--  1 fgv 1.3M May  2 09:40 voldemort-1.10.14-src.jar
-rw-r--r--  1 fgv 311K May  2 09:40 voldemort-contrib-1.10.14.jar
-rw-r--r--  1 fgv 1.2M May  2 09:40 voldemort-test-1.10.14.jar
[fgv-ld1:voldemort]$ ./bin/voldemort-shell.sh 
Usage: voldemort-shell.sh store_name bootstrap_url [command_file] [options]
Option                               Description                           
------                               -----------                           
--client-zone-id <Integer: zone-id>  Client zone id for zone routing       
--config-file <file>                 A properties file that contains client
                                       config properties                   
--help                               Print this help message               
--voldemort-shell <script_suffix>    Suffix of shell script; used to format
                                       help output. Examples of script     
                                       suffixes: sh, bat, app              
```

We're going to need to tweak the new Gradle build or revert this change. Hopefully the former...

-F

FelixGV:
Ok nevermind...

@arunthirupathi told me that the end result of ```./gradlew jar``` before this change can now be achieved with ```./gradlew build -x test```.

-F

tempredirect:
'assemble' should probably work, if it doesn't then it shouldn't be to tricky to hook up what ever task copies all the libs into place.

Gareth

arunthirupathi:
For people who are used to jar task, the new jar task completes early, so
it will not have the required functionality.

We updated the docs to use
./gradlew build -x test  # Run build, but skip all the test tasks.

But still there could be issues around the new multi module build, please
let us know and we will be happy to help.

Thanks,
Arun.

PR #352, opened 2015-12-03
Let read only server use bouncy castle as JCE provider.

squarY:
Add a property in configuration "readonly.bouncy.castle.enabled" to let voldemort read-only server enable bouncy castle as JCE provider.

### DIFF
--- a/src/java/voldemort/server/VoldemortConfig.java
+++ b/src/java/voldemort/server/VoldemortConfig.java
@@ -180,6 +180,7 @@ public class VoldemortConfig implements Serializable {
     private long defaultStorageSpaceQuotaInKB;
     private String modifiedProtocol;
     private int modifiedPort;
+    private boolean bouncyCastleEnabled;
 
     public static final String PUSH_HA_ENABLED = "push.ha.enabled";
     private boolean highAvailabilityPushEnabled;
@@ -406,6 +407,7 @@ public class VoldemortConfig implements Serializable {
 
         this.modifiedProtocol = props.getString("readonly.modify.protocol", null);
         this.modifiedPort = props.getInt("readonly.modify.port", -1);
+        this.bouncyCastleEnabled = props.getBoolean("use.bouncycastle.for.ssl", false);
 
         this.highAvailabilityPushClusterId = props.getString(PUSH_HA_CLUSTER_ID, null);
         this.highAvailabilityPushLockPath = props.getString(PUSH_HA_LOCK_PATH, null);
@@ -3051,6 +3053,21 @@ public class VoldemortConfig implements Serializable {
         this.modifiedPort = modifiedPort;
     }
 
+    public boolean isBouncyCastleEnabled () {
+        return bouncyCastleEnabled;
+    }
+
+    /**
+     * Set whether use bouncy castle as JCE provider or not.
+     *
+     * <ul>
+     * <li>Property : "use.bouncycastle.for.ssl"</li>
+     * <li>Default : false</li>
+     * </ul>
+     */
+    public void setBouncyCastleEnabled (boolean bouncyCastleEnabled) {
+        this.bouncyCastleEnabled = bouncyCastleEnabled;
+    }
     public boolean isHighAvailabilityPushEnabled() {
         return highAvailabilityPushEnabled;
     }
--- a/src/java/voldemort/server/VoldemortServer.java
+++ b/src/java/voldemort/server/VoldemortServer.java
@@ -21,12 +21,14 @@ import static voldemort.utils.Utils.croak;
 import java.io.File;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.security.Security;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 
 import org.apache.log4j.Logger;
 
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import voldemort.VoldemortException;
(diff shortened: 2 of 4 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -76,16 +78,11 @@
...
-             config.getReadOnlyFetcherMaxBytesPerSecond(),
-             config.getReadOnlyFetcherReportingIntervalBytes(),
-             config.getReadOnlyFetcherThrottlerInterval(),
-             config.getFetcherBufferSize(),
-             config.getReadOnlyFetchRetryCount(),
-             config.getReadOnlyFetchRetryDelayMs(),
-             config.isReadOnlyStatsFileEnabled(),
-             config.getReadOnlyMaxVersionsStatsFile(),
-             config.getFetcherSocketTimeout());
+        this(config, config.getReadOnlyFetcherMaxBytesPerSecond(), config.getReadOnlyFetcherReportingIntervalBytes(),
```
FelixGV: Can we try to avoid unnecessary formatting changes please? I can export my IntelliJ editor settings if you'd like (:

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -220,6 +217,12 @@ private File fetchFromSource(String sourceFileUrl,
         ObjectName jmxName = null;
         HdfsCopyStats stats = null;
         FileSystem fs = null;
+
+        if (voldemortConfig.isBouncyCastleEnabled()) {
+            //Use bouncy castle provider as first choice. If provider is existed, it will not be inserted again.
```
FelixGV: Can you clarify, what do you mean when you say "If provider is existed"? Perhaps "If the BouncyCastleProvider was already inserted"?

FelixGV:
Left two minor comments. LGTM! Thanks (:

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -406,6 +407,7 @@ public VoldemortConfig(Props props) {
 
         this.modifiedProtocol = props.getString("readonly.modify.protocol", null);
         this.modifiedPort = props.getInt("readonly.modify.port", -1);
+        this.bouncyCastleEnabled = props.getBoolean("readonly.bouncy.castle.enabled", false);
```
arunthirupathi: I would call this parameter,  use.bouncycastle.for.ssl and initialize this here itself.  As the way it is done in HdfsFetcher it impacts all the SSL from that point on, so it would be safer to do it this way I believe.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
squarY: Will fix the comment. Yes, if BouncyCastleProvider had been inserted, this insert method will return instead of insert provider again.

src/java/voldemort/server/VoldemortConfig.java
squarY: I thought about the option you said. If we'd like to use BouncyCastleProvider for all SSL encryption/decryption, yes, we should initialize here. Will fix it.

FelixGV: I'm not sure it's a good idea to alter the state of the JVM from within the config class. I would rather leave that in the HdfsFetcher for now. And if/when we want to introduce SSL in Voldemort, we can move that initialization code elsewhere, if need be.

I'm fine with either name. Your call.

arunthirupathi: Probably not in the config class, but some place common. Leaving it in the HDFSFetcher alters the behavior of the program mid way and it will have inconsistent behavior for some paths. It need to be done in the startup so the behavior is consistent, let me look for a place to put this in.

FelixGV: All right, sounds good.

arunthirupathi: VoldemortServer constructor seems like a good place. 

There are two constructors, one for test and other for actual flow, so you need to put it at both the places.

squarY: Done

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
squarY: Fixed. Yes, please send me the code format file, thanks~

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -140,6 +144,16 @@ public VoldemortServer(VoldemortConfig config, Cluster cluster) {
         createOnlineServices();
     }
 
+    private void enableBouncyCastleForSSL(boolean enable) {
+        if (enable) {
+            if (voldemortConfig.isBouncyCastleEnabled()) {
```
arunthirupathi: Can we remove the enable parameter for this method, the same config is read twice.

squarY: Done.

arunthirupathi:
looks good to me, squash and merge it

PR #353, opened 2015-12-04
BnP logging improvements

FelixGV:
- BnP job now emits config properties in one entry per line.
- Bumped up some useful logs to INFO level in the HttpHook.


FelixGV:
Thanks for the review, I merged it in.

PR #357, opened 2015-12-14
BouncyCastle jar is required even when not enabled

arunthirupathi:
For some background read this post
http://stackoverflow.com/questions/34259275/when-is-a-java-class-loaded/

This command failed before and after this patch it passed.
./gradlew clean jar && rm lib/bcprov-jdk15on-1.48.jar && bin/voldemort-server.sh config/single_node_cluster/

BouncyCastle jar is referenced from VoldemortServer and when
VoldemortServer class is loaded, all its references are resolved
which causes it to fail with ClassLoader error.

Added one more level of indirection to avoid the direct reference
and hence BouncyCastle is not required to be available in the class
path when not enabled.

### DIFF
--- a/src/java/voldemort/server/SetupSSLProvider.java
+++ b/src/java/voldemort/server/SetupSSLProvider.java
@@ -1,0 +1,11 @@
+package voldemort.server;
+
+import java.security.Security;
+
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+
+public class SetupSSLProvider {
+    public static void useBouncyCastle() {
+        Security.insertProviderAt(new BouncyCastleProvider(), 1);
+    }
+}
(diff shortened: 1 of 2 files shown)


FelixGV:
That should be fine... But while we're at it, instead of having a config which is "bouncy castle = true|false", which not make it fully generic? The config could be ```security.provider=fully.qualified.class.name```, and then we'd load that class via reflection, and it's the operator's responsibility to inject whatever they need on the classpath?

-F

arunthirupathi:
Felix, though I think that is very flexible I don't see a need for it. There are not that many crypto providers except for JCE and BouncyCastle. Implementing a crypto provider means you need to have your jar signed by Oracle.  So we will save this until required.

singhsiddharth:
+1 to what Arun said. Let's keep it simple for now.

FelixGV:
Sure...

FelixGV:
I merged this in.

PR #359, opened 2015-12-16
Read-Only fetches now abort immediately when killed.

FelixGV:
Also improved vadmin.sh's output format and error handling.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
@@ -16,8 +16,10 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.log4j.Logger;
 
 import voldemort.server.protocol.admin.AsyncOperationStatus;
+import voldemort.server.protocol.admin.AsyncOperationStoppedException;
 import voldemort.store.readonly.checksum.CheckSum;
 import voldemort.store.readonly.checksum.CheckSum.CheckSumType;
+import voldemort.utils.ExceptionUtils;
 
 
 public class BasicFetchStrategy implements FetchStrategy {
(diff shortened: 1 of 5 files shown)


src/java/voldemort/tools/admin/command/AdminCommandAsyncJob.java
```diff
@@ -182,12 +183,18 @@ public static void executeCommand(String[] args) throws IOException {
...
+                System.out.println(" : " + asyncIds);
                 for(int asyncId: asyncIds) {
-                    System.out.println("Async Job Id " + asyncId + " ] "
-                                       + adminClient.rpcOps.getAsyncRequestStatus(nodeId, asyncId));
+                    System.out.print("\tAsync Job Id " + asyncId + " : ");
+                    try {
+                        AsyncOperationStatus status = adminClient.rpcOps.getAsyncRequestStatus(nodeId, asyncId);
+                        System.out.println(status);
+                    } catch (Exception e) {
+                        System.out.println("threw " + e.getClass().getSimpleName() + "!");
```
arunthirupathi: Can we log the message too, I believe only class name is logged here.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
```diff
@@ -120,6 +122,14 @@ public BasicFetchStrategy(HdfsFetcher fetcher,
                 int read;
 
                 while (true) {
+                    if (status.hasException()) {
+                        Exception ex = status.getException();
+                        if (ex.getClass().equals(AsyncOperationStoppedException.class)) {
```
arunthirupathi: should it be instance of AsyncOperationStoppedException instead of class name equals ?

arunthirupathi:
Left some comments, looks good to me.

src/java/voldemort/tools/admin/command/AdminCommandAsyncJob.java
FelixGV: The message is included as part of the stacktrace, which is printed on the line below.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
FelixGV: Yeah, I can change that.

FelixGV:
Thanks for the review, I addressed your comment, rebased and merged it in (:

PR #360, opened 2015-12-17
Add a tool to compare and get diff records for readonly store.

squarY:
Add a tool ReadOnlyFileDiffer to compare between the two versions of readonly data and print the number of added, deleted, updated and same records.


FelixGV:
Very cool, I'll review ASAP.

bin/read-only-file-differ.sh
```diff
@@ -0,0 +1,11 @@
...
+
+base_dir=$(cd $(dirname $0)/.. && pwd)
+
+VERSION=`grep 'curr.release' $base_dir/gradle.properties | sed -e 's/curr.release=\(.*\)/\1/'`
+echo "Voldemort version detected: $VERSION"
+
+# The jar file's name depends on the directory name...
+export VOLDEMORT_JAR="$(echo $base_dir/build/libs/voldemort-$VERSION.jar)"
+
+java -cp $VOLDEMORT_JAR:$base_dir/lib/* voldemort.store.readonly.ReadOnlyFileDiffer $@
```
arunthirupathi: Did you take a look at the voldemort-shell.sh ? We have bin/run-class.sh for the same purpose. I don't see dependent jars in the class path.

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+        try {
+            cmd = parser.parse(options, args);
+            //Print help information.
+            if (cmd.hasOption("h")) {
+                HelpFormatter formatter = new HelpFormatter();
+                formatter.printHelp("ReadOnlyFileDiffer", options);
+                return;
+            }
+
+            String path = cmd.getOptionValue("p");
```
arunthirupathi: Can you make sure that if the parameters are not passed in correctly, it will print the help message and error out ? AFAIK, Options setRequired ensures that you are parameters are not flags and requires value, but you can skip the entire parameter and it does not notify you.

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+                "Start comparing for store:" + store + " between version-" + oldVersion + " and version-" + newVersion);
+
+            long[] diffResults = analyzer.compareForStore(path, store, oldVersion, newVersion, rate);
+
+            System.out.println(
+                "Diff of store" + store + ". Added:" + diffResults[ADD] + " Deleted:" + diffResults[DELETE]
+                    + " Updated:" + diffResults[UPDATE] + " Same:" + diffResults[SAME]);
+        } catch (ParseException e) {
+            System.err.println(e.getMessage());
+            System.exit(-1);
```
arunthirupathi: Can you also print the exception stack trace in case of error.

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    static class ByteArrayComparator implements Comparator<byte[]> {
+
+        @Override
+        public int compare(byte[] o1, byte[] o2) {
+            int i = 0;
+            int j = 0;
+            for (; i < o1.length && j < o2.length; i++, j++) {
+                if (o1[i] < o2[i]) {
```
arunthirupathi: AFAIK, java's byte is signed type. Ranging from -128 to 127. I don't know whether it would cause problem in comparison. But I would prefer the byte array comparator to use one of the proven methods instead of we writing one from scratch. 

http://stackoverflow.com/questions/5108091/java-comparator-for-byte-array-lexicographic

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    /**
+     * This class is used to compare between chunk files. Smaller chunk number means smaller when comparing.
+     */
+    static class ReadOnlyFileNameComparator implements Comparator<String> {
+
+        @Override
+        public int compare(String o1, String o2) {
+            int chunk1 = Integer.valueOf(o1.substring(o1.lastIndexOf('_') + 1));
```
arunthirupathi: Why is +1 added to both the numbers and then subtracted ?

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+            int chunk2 = Integer.valueOf(o2.substring(o2.lastIndexOf('_') + 1));
+            return chunk1 - chunk2;
+        }
+    }
+
+    /**
+     * Iterator of records in one partition. No matter how many chunk files in this partition,
+     * iterator could get next record correctly in this partition.
+     */
+    static class ReadOnlyRecordIterator implements Iterator<byte[]> {
```
arunthirupathi: I would prefer this to be in a class file by itself.

arunthirupathi:
Yan, I left some comments, but the code quality is very good, loved it. It was easy to read and understand and very modular. They are all very minor comments.

Since this is all new code and does not touch any existing code path, there is no impact to the existing scenarios.

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+import voldemort.utils.ByteUtils;
+import voldemort.utils.EventThrottler;
+import voldemort.utils.Pair;
+
+
+/**
+ * ReadOnlyFileDiffer is used to compare between different versions of specific store and give the number of how many
+ * records are added/deleted/updated/same.
+ *
+ * N.B.: This tool only works with the Read-Only V2 format.
```
FelixGV: Nitpick, since you are inside a JavaDoc block, you can use the following notation:

{@link voldemort.store.readonly.ReadOnlyStorageFormat#READONLY_V2}

This creates a link in auto-generated JavaDoc and also allows you to follow the link using IntelliJ. Not really important, but I just thought I'd mention it (:

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
```diff
@@ -0,0 +1,383 @@
...
+        File oldVersionDir = getVersionDir(path, store, oldVersion);
+        File newVersionDir = getVersionDir(path, store, newVersion);
+        HashSet<String> OldPartitions = getPartitionPrefixes(oldVersionDir);
+        HashSet<String> newPartitions = getPartitionPrefixes(newVersionDir);
+        ArrayList<String> comparablePartitions = new ArrayList<String>();
+        ByteArrayComparator comparator = new ByteArrayComparator();
+
+        //Only compare common partitions which exist in both versions.
+        for (String partition : OldPartitions) {
+            if (newPartitions.contains(partition)) {
```
FelixGV: You should probably log an error if that is not the case...

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
FelixGV: The +1 is to tell the ```substring``` function to start its selection at the index which immediately follows the index of the last underscore. In other words, that is the chunk ID in the file name.

FelixGV:
Very nice code! Left some minor comments.

src/java/voldemort/store/readonly/ReadOnlyFileDiffer.java
lennyxue: why print log by system out rather than by logger? there is no time info printed by console

PR #361, opened 2015-12-17
Made vadmin.sh more resilient to node failures.

FelixGV:


src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -2016,67 +2016,79 @@ private String diffMessage(StoreDefinition newStoreDef, StoreDefinition remoteSt
...
+         * @throws VoldemortException of the first node which failed (note, there might be more)
+         * @see {@link #deleteStore(String, java.util.List)} for more visibility into specific failures.
          */
         public void deleteStore(String storeName) {
-            for(Node node: currentCluster.getNodes()) {
-                deleteStore(storeName, node.getId());
+            List<Integer> nodeIds = Lists.newArrayList(currentCluster.getNodeIds());
+            Map<Integer, VoldemortException> exceptionMap = deleteStore(storeName, nodeIds);
+            if (!exceptionMap.isEmpty()) {
+                throw exceptionMap.values().iterator().next();
```
arunthirupathi: This throws exception from one of the nodes. I would prefer the error message to be Deleting stores failied on the following nodes x,y,z and with one of the exceptions.

FelixGV: Sure, I could do that, but it's possible that there are different types of failures (i.e.: store doesn't exist on some nodes, while some other node is unreachable). What I did here is to maintain the previous behavior (throwing the "first" exception that happens), except that now there is a guarantee that it will have attempted the delete on every node, instead of stopping after the first failure.

arunthirupathi: I see, this sounds fine to me, Please merge this in.

arunthirupathi:
@FelixGV  do you want to merge this change in, this change looks good to me.

FelixGV:
Yeah I'll merge it in later. Or you can do it if you prefer.

arunthirupathi:
I will merge this in later, about to release a new version now.

FelixGV:
I pushed this one in.

PR #362, opened 2015-12-18
Ro file length

arunthirupathi:
Added validation of RO files to vadmin.sh meta check command.

When RO errors are detected, it is going to be verbose and it will take time to analyze the errors. But at least we will be aware of the errors, and can take preventive actions.

It warns for following reasons.
1) RO Store with replication factor of 1.
2) RO Store with RF > 1, and sizes not match for same partition, chunk.
3) RO store with RF > 1, file missing from one of the nodes.


FelixGV:
LGTM! Thanks (: !

PR #363, opened 2015-12-18
Disabling Fetch Fails BnP HA

arunthirupathi:
When a node is in offline mode, it responds with fetch disabled
error. Currently the error is not treated as a soft error and this
fails the HA BnP.

With this code change, FetchDisabled is considered as a soft error
and the fetch will continue as normal.


FelixGV:
Both changes LGTM!

PR #365, opened 2016-01-06
Fix for a regression introduced in a706147

FelixGV:
Under certain partition assignment configurations, servers would skip the
initialization of certain partitions, which would then result in NPEs getting
thrown during get requests. This issue is now fixed.

In addition to that, the following changes are also included in this commit:
- More comprehensive tests in ReadOnlyStorageEngineTest to catch the edge case.
- Better error reporting in ChunkedFileSet.getChunkForKey(byte[] key).

Currently re-running all unit tests. Will report back when that's finished.

### DIFF
--- a/src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
+++ b/src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
@@ -519,8 +519,7 @@ public class ReadOnlyStorageEngine extends AbstractStorageEngine<ByteArray, byte
             fileModificationLock.readLock().lock();
             int chunk = fileSet.getChunkForKey(key.get());
             if(chunk < 0) {
-                logger.warn("Invalid chunk id returned. Either routing strategy is inconsistent or storage format not understood");
-                return Collections.emptyList();
+                throw new IllegalStateException("chunkId < 0"); // should never happen, but just in case
             }
             int location = searchStrategy.indexOf(fileSet.indexFileFor(chunk),
                                                   fileSet.keyToStorageFormat(key.get()),
@@ -535,6 +534,10 @@ public class ReadOnlyStorageEngine extends AbstractStorageEngine<ByteArray, byte
             } else {
                 return Collections.emptyList();
             }
+        } catch (IllegalStateException e) {
+            logger.warn("ChunkedFileSet.getChunkForKey() did not execute successfully. " +
+                        "Returning empty result for key: " + ByteUtils.toHexString(key.get()), e);
+            return Collections.emptyList();
         } finally {
             fileModificationLock.readLock().unlock();
         }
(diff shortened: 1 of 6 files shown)


src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -535,6 +534,10 @@ public void truncate() {
             } else {
                 return Collections.emptyList();
             }
+        } catch (IllegalStateException e) {
+            logger.warn("ChunkedFileSet.getChunkForKey() did not execute successfully. " +
+                        "Returning empty result for key: " + ByteUtils.toHexString(key.get()), e);
```
arunthirupathi: Can you also include the storeName in this message ?

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -551,16 +554,21 @@ public void truncate() {
...
+                    if(chunk < 0) {
+                        throw new IllegalStateException("chunkId < 0"); // should never happen, but just in case
+                    }
+                    int valueLocation = searchStrategy.indexOf(fileSet.indexFileFor(chunk),
+                                                               fileSet.keyToStorageFormat(key.get()),
+                                                               fileSet.getIndexFileSize(chunk));
+                    if(valueLocation >= 0)
+                        keysAndValueLocations.add(new KeyValueLocation(chunk, key, valueLocation));
+                } catch (IllegalStateException e) {
+                    logger.warn("ChunkedFileSet.getChunkForKey() did not execute successfully. " +
```
arunthirupathi: Same here.

src/java/voldemort/store/readonly/StoreVersionManager.java
```diff
@@ -78,7 +78,7 @@ public void syncInternalStateFromFileSystem() {
         // Make sure versions missing from the file-system are cleaned up from the internal state
         for (Long version: versionToEnabledMap.keySet()) {
             File[] existingVersionDirs = ReadOnlyUtils.getVersionDirs(rootDir, version, version);
-            if (existingVersionDirs.length == 0) {
+            if (existingVersionDirs == null || existingVersionDirs.length == 0) {
```
arunthirupathi: This doesn't feel right to me, ReadOnlyUtils.getVersionDirs should return an empty list. There are too many callers, if there is a semantic change introduced, I would expect it to return an empty list instead of handling it all the callers.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
FelixGV: Yeah totally. I actually thought about doing that and then got distracted. I'll add it in.

src/java/voldemort/store/readonly/StoreVersionManager.java
FelixGV: Yeah, that's not important to this fix. I can remove it.

FelixGV:
Thanks a lot for the quick review guys.

I addressed your comments, cleaned up a few things, rebased, squashed and pushed to master.

PR #366, opened 2016-01-06
AdminClient does not obey the Client Timeout

arunthirupathi:
Review the changes in AdminClient and Azkaban* files. Others can be ignored. 

1) For BuildAndPush If the Voldemort Cluster is far away from the
Hadoop Cluster, the connection timeout causes the Job to fail.
Increased the connection timeout in Azkaban.* files for this issue.

2) ClientConfig timeout is ignored by the AdminClient bootstrap
methods and it constructs an arbitary ClientConfig. Fixed that.

3) AdminClient passes in empty AdminClientConfig and ClientConfig
at multiple places. Removed that.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortRollbackJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortRollbackJob.java
@@ -27,9 +27,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.log4j.Logger;
 
 import voldemort.VoldemortException;
-import voldemort.client.ClientConfig;
 import voldemort.client.protocol.admin.AdminClient;
-import voldemort.client.protocol.admin.AdminClientConfig;
 import voldemort.cluster.Cluster;
 import voldemort.cluster.Node;
 import voldemort.store.readonly.mr.utils.VoldemortUtils;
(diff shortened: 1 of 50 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortRollbackJob.java
```diff
@@ -76,9 +74,7 @@ public void run() throws Exception {
             ExecutorService service = null;
             try {
                 service = Executors.newCachedThreadPool();
-                adminClient = new AdminClient(clusterUrl,
-                                              new AdminClientConfig(),
-                                              new ClientConfig());
+                adminClient = new AdminClient(clusterUrl);
```
FelixGV: This one needs a higher timeout as well, like you did in ```VoldemortBuildAndPushJob```, right?

arunthirupathi: I searched for the callers of the RollBackJob and there is no reference to it. There is not even a main method which is weird. Will look more into it.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -264,38 +276,12 @@ public AdminClient(String bootstrapURL,
...
-                       int zoneID) {
-        this(bootstrapURL, adminClientConfig, clientConfig);
-        helperOps.initSystemStoreClient(bootstrapURL, zoneID);
+        this(adminClientConfig, clientConfig, cluster);
     }
 
     @Override
     public String toString() {
-        return "AdminClient with mainBootstrapUrl: " + this.mainBootstrapUrl;
+        return debugInfo;
```
FelixGV: Hanging on to this looks a bit weird, but I guess that's ok.

FelixGV:
LGTM.

PR #367, opened 2016-01-08
Fixed a bug where the Read-Only data directory config is not honored properly.

FelixGV:
This fixes a regression introduced in a7061474f905fe60a9581dbf4e14431d00b130a2.


arunthirupathi:
LGTM, I believe we need to compare some old and new configs to see if the config is the same after the refactoring. Will try to do that today.

PR #368, opened 2016-01-08
Bug fixes and improvements for Build and Push.

FelixGV:
- HadoopStoreWriter did not handle multiple chunks correctly which caused reduce tasks to step on each others' toes. Fixed.
- Changed default 'reducer.per.bucket' config to true.
- AbstractStoreBuilderConfigurable.getPartition() was broken when 'build.primary.replicas.only' was enabled. Fixed it and added a lot more comments.
- Added defensive code in HadoopStoreWriter to catch future regressions in the partitioning (shuffling) code.
- Improved logging in Props and HadoopStoreBuilder.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
@@ -22,8 +22,10 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.Locale;
+import java.util.Set;
 import java.util.zip.GZIPOutputStream;
 
+import com.google.common.collect.Sets;
 import org.apache.commons.codec.binary.Hex;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
@@ -36,6 +38,7 @@ import org.apache.hadoop.mapred.Reporter;
 import org.apache.log4j.Logger;
 
 import voldemort.VoldemortException;
+import voldemort.annotations.concurrency.NotThreadsafe;
 import voldemort.store.readonly.ReadOnlyStorageMetadata;
 import voldemort.store.readonly.ReadOnlyUtils;
 import voldemort.store.readonly.checksum.CheckSum;
@@ -51,6 +54,8 @@ public class HadoopStoreWriter
         implements KeyValueWriter<BytesWritable, BytesWritable> {
 
     private static final Logger logger = Logger.getLogger(HadoopStoreWriter.class);
+
+    private Set<Integer> chunksHandled = Sets.newHashSet();
 
     private DataOutputStream[] indexFileStream = null;
     private DataOutputStream[] valueFileStream = null;
@@ -73,12 +78,14 @@ public class HadoopStoreWriter
 
     private FileSystem fs;
 
+    private boolean isValidCompressionEnabled;
+
     /**
      * This variable is used to figure out the file extension for index and data
      * files. When the server supports compression, this variable's value is
      * typically ".gz" or else it holds and empty string
      */
-    private String fileExtension = null;
+    private String fileExtension;
 
     /** Used to track the size of the index files produced by the reducer */
     private long[] indexFileSizeInBytes;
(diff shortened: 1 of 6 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
```diff
@@ -213,13 +226,33 @@ public void write(BytesWritable key, Iterator<BytesWritable> iterator, Reporter
...
+            }
             offsetTillNow += ByteUtils.SIZE_OF_INT;
 
             // Read partition id
-            if(this.partitionId == -1)
-                this.partitionId = ByteUtils.readInt(valueBytes, offsetTillNow);
+            int currentPartitionId = ByteUtils.readInt(valueBytes, offsetTillNow);
+            if (this.partitionId == -1) {
+                this.partitionId = currentPartitionId;
+            } else if (this.partitionId != currentPartitionId) {
```
arunthirupathi: we should also validate the replica type. The error message sounds little confusing.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
```diff
@@ -132,56 +130,69 @@ private void initFileStreams(boolean isCompressionEnabled, String compressionCod
...
+    /**
+     * The MapReduce framework should operate sequentially, so thread safety shouldn't be a problem.
+     */
+    @NotThreadsafe
+    private void initFileStreams(int chunkId) {
+        /**
+         * {@link Set#add(Object)} returns false if the element already existed in the set.
+         * This ensures we initialize the resources for each chunk only once.
+         */
+        if (chunksHandled.add(chunkId)) {
```
arunthirupathi: I believe this can introduce regression. Let us say that you have chunks 1 to 5 and when we partition the key space we only got 1,2,3 and 5 ( 4 goes missing). The way our RO code is initialized it will read 1,2,3 and will never read 5. 

Though I like what you are doing here, I would suggest keeping the regression fix to be barely minimal. We will introduce refactoring, immediately after that. Valuable lesson I learnt when we work on hot fixes before the final build.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
FelixGV: Yeah the message is bad. I meant to write: "Should NOT get various partitionId shuffled to us!"

The replica type is validated below.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
FelixGV: Hmm I think I see what you mean. I might have a way to reproduce the bug in your theory. Will talk to you when I get in.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
arunthirupathi: The replica type is validated for the first read only for BPRO. Subsequent replica types are not validated.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
arunthirupathi: It is probably going to very hard to reproduce unless you have crafted data. But nevertheless I believe that is an issue.

FelixGV: I can artificially increase the number of chunks via config, and for sure some of them will be empty.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
FelixGV: Yeah ok, I can add validation on every record.

PR #369, opened 2016-01-08
fixing the 'unknown.host' test in ClientRequestExecutorPoolTest

tempredirect:
it turns out that somebody has registered 'unknown.host':

    $ host unknown.host
    unknown.host has address 188.68.51.215
    unknown.host mail is handled by 10 serpens.uberspace.de.

which is kinda of annoying. Have switched to using unknown.invalid as
this is reserved and should never resolve.


arunthirupathi:
I pulled your branch and manually merged it, to keep the version history clean. Thanks for the contribution.

arunthirupathi:
This is your commit https://github.com/voldemort/voldemort/commit/b46c97faf9a079961cd8b3dc701138b1ffc8c9dd

tempredirect:
np

PR #371, opened 2016-01-12
Changed HDFS directory size measurement in HdfsFetcher.

FelixGV:
This fixes two issues in 'build.primary.replicas.only':
1. The progress report in the AsyncTask was wrong, which
   also resulted in erroneous logs in the BnP job.
2. The previous implementation was DDOSing the NameNode
   with a furious amount of ListStatus operation, most
   of which it did not even need to get the result of.

Also cleaned up HdfsFetcherAdvancedTest a little bit.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
@@ -83,7 +83,7 @@ public class HdfsCopyStats {
         return this.statsFile;
     }
 
-    private static final int STATS_VERSION = 3;
+    private static final int STATS_VERSION = 4;
     private void initializeStatsFile(File destination,
                                      boolean enableStatsFile,
                                      int maxVersionsStatsFile,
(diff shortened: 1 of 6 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
```diff
@@ -230,16 +230,23 @@ public void reportFileDownloaded(File file,
...
+                    + " . Total bytes written: " + totalBytesWritten
+                    + " . Time taken(MS): " + (nowMS - startTimeMS));
         if(statsFileWriter != null) {
             IOUtils.closeQuietly(statsFileWriter);
         }
     }
 
     public double getPercentCopied() {
+        if (pathInfo == null) {
+            return -1;
```
singhsiddharth: -1.0 ?

FelixGV: Sure. Will do.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsDirectory.java
```diff
@@ -50,15 +55,52 @@ public HdfsFile getMetadataFile() {
...
+            }
+        }
+
         if(files == null) {
             throw new VoldemortApplicationException(source + " is empty");
         }
 
         for(FileStatus file: files) {
+            // Bookkeeping
+            totalSizeOfChildren += file.getLen();
```
arunthirupathi: Should this be done only for files ?

arunthirupathi:
Looks good.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsDirectory.java
FelixGV: Hmm... I'm not sure what gets reported by ```getLen()``` when the ```FileStatus``` instance represents a directory. The JavaDoc says ```@return the length of this file, in bytes.``` but it does not mention what is returned for directories.

arunthirupathi: It should not be an issue for now , as we don't have subdirectories and the code path is invalid, we will clean that up later.

arunthirupathi:
Since you have merged, probably you can close this Pull request ?

FelixGV:
Oh yeah, I forgot. Thanks.

PR #372, opened 2016-01-19
Better exception handling in HdfsFetcher#fetchFromSource().

FelixGV:
Previously, a FileSystem#close() operation was guarded by try/catch looking for IOExceptions only. It turns out this close() function can sometimes throw NPEs as well. This commit expands the catch clause to any exception type, and updates the logged message accordingly.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -368,12 +368,13 @@ public class HdfsFetcher implements FileFetcher {
             if(fs != null) {
                 try {
                     fs.close();
-                } catch(IOException e) {
-                    String errorMessage = "Got IOException while trying to close the filesystem instance (harmless).";
+                } catch(Exception e) {
+                    String errorMessage = "Caught " + e.getClass().getSimpleName() +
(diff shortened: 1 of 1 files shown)


arunthirupathi:
I don't want it to be error, probably I was thinking about logger.debug ?

Also I am concerned little that, if fs.close fails, can it mean some underlying filesystem error ?

FelixGV:
Well, just to be clear, the file-system in question is a Hadoop file-system instance, and it used for fetching data, so it would seem pretty weird that a read error would wait until close-time to be bubbled up. If we were writing to the file-system, that may be more worrisome, but we are not in this case. That being said, we've seen plenty of weird things happening with Hadoop file-system instances, so who knows xD

FelixGV:
But since we are already logging the error in the stats file, I agree we could forego putting it at INFO-level in the server logs as well.

FelixGV:
I pushed an amended commit which lowers the logging level to DEBUG, @arunthirupathi.

arunthirupathi:
Ah..... I see, I thought the file that we are writing to the disk. I am not that concerned about the hadoop fs close as our checksum should catch it hopefully :)

FelixGV:
Yeah, indeed.

FelixGV:
So... shall I merge it in (: ?

arunthirupathi:
Yes, Please do.

Thanks,
Arun.

FelixGV:
Thanks for the review @arunthirupathi

PR #373, opened 2016-01-22
Fix exception error in ClientRequestExecutorPoolTest

SidW:
The test failed because the original code in order to establish socket connection throws out 'UnresolvedAddressException' instead of 'ConnectException'. So I updated test cases.

### DIFF
--- a/test/unit/voldemort/server/socket/ClientRequestExecutorPoolTest.java
+++ b/test/unit/voldemort/server/socket/ClientRequestExecutorPoolTest.java
@@ -289,14 +289,7 @@ public class ClientRequestExecutorPoolTest {
         SocketDestination nonExistentHost = new SocketDestination("unknown.invalid",
                                                                   port,
                                                                   RequestFormatType.VOLDEMORT_V1);
-        // JDK 1.6 throws UnresolvedAddressException, which is not even an
-        // instance of IOException or ConnectException . JDK 1.8 fixes this
-        // issue and uses the ConnectException .
-        try {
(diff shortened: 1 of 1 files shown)


test/unit/voldemort/server/socket/ClientRequestExecutorPoolTest.java
```diff
@@ -292,11 +292,7 @@ public void testNonExistentHost() throws Exception {
         // JDK 1.6 throws UnresolvedAddressException, which is not even an
         // instance of IOException or ConnectException . JDK 1.8 fixes this
         // issue and uses the ConnectException .
-        try {
-            testConnectionFailure(pool, nonExistentHost, ConnectException.class);
-        } catch(Exception e) {
-            testConnectionFailure(pool, nonExistentHost, UnresolvedAddressException.class);
-        }
+        testConnectionFailure(pool, nonExistentHost, UnresolvedAddressException.class);
```
arunthirupathi: Sidian, can you also remove the comments above it. I was stupid https://github.com/voldemort/voldemort/commit/fa5bf0e2356de94edfa11d18cd7602fe0f78a29b

The exception is same in JDK 1.6 and 1.8 , but some one registered the old address which caused the exception to be different. 

This fixed it, but my regression broke it.
https://github.com/voldemort/voldemort/commit/b46c97faf9a079961cd8b3dc701138b1ffc8c9dd

arunthirupathi:
Cool.... can you please squash all this into single commit, this will help the master branch history to stay clean ?

PR #375, opened 2016-02-09
Generate a new fat jar which solves the dependency conflict issues in some cases. 

SidW:
Both protoBuf and sleepyCat dependencies are now integrated into the fat jar. 
Relocated the paths in case that there would be dependency conflict issue occurred.


arunthirupathi:
I merged this with ammended commit message.

https://github.com/voldemort/voldemort/commit/abcf36a7c7fffdfd773d1131be93b946ec3923e4

PR #377, opened 2016-02-26
Stop swallowing exception's original message and clarify top-level message

SidW:
AdminStoreSwapper#invokeFetch() used to take charge of collecting exceptions among sub-threads. It swallowed original exceptions and made customers confused when debugging. Now it maintains all sub-thread exception in an list and throw the first exception in the list.

Also modify the InvalidBootstrapurLException message to make it more clear. It is now:
 "Not able to find store (" + storeName +") in this cluster according to the push URL. BnP job is not able to create new stores now." + "Please create new stores through Nuage if you think this is the correct cluster you want to push."

### DIFF
--- a/src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
+++ b/src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
@@ -213,8 +213,9 @@ public class AdminStoreSwapper {
          * that this needs careful decision on how to handle those fetches that
          * already started in other nodes and how & when to clean them up.
          */
-        int numQuotaExceededException = 0;
-        boolean invalidBootstrapURLExceptions = false;
+        /*int numQuotaExceededException = 0;
+        ArrayList<Integer>QuotaExceededExceptions = new ArrayList<Integer>();
+        //boolean invalidBootstrapURLExceptions = false;
         for(final Node node: cluster.getNodes()) {
             Future<String> val = fetchDirs.get(node.getId());
             try {
@@ -223,10 +224,14 @@ public class AdminStoreSwapper {
             } catch(ExecutionException e) {
                 fetchErrors = true;
                 if(e.getCause() instanceof QuotaExceededException) {
-                    numQuotaExceededException++;
+                    //numQuotaExceededException++;
+                    QuotaExceededExceptions.add(node.getId());
                     fetchResponseMap.put(node, new Response((QuotaExceededException) e.getCause()));
                 } else if(e.getCause() instanceof InvalidBootstrapURLException) {
-                    invalidBootstrapURLExceptions = true;
+                    //invalidBootstrapURLExceptions = true;
+                    throw new InvalidBootstrapURLException("Not able to find store (" + storeName +
+                            ") in this cluster according to the push URL. BnP job is not able to create new stores now." +
+                            "Please create new stores through Nuage if you think this is the correct cluster you want to push." );
                 } else {
                     fetchResponseMap.put(node, new Response(e));
                 }
@@ -238,9 +243,29 @@ public class AdminStoreSwapper {
         }
 
         // Invalid stores should fail faster
-        if(invalidBootstrapURLExceptions) {
-            throw new InvalidBootstrapURLException("Exceptions during push. Invalid bootstrap url. Please check your "
+        /*if(invalidBootstrapURLExceptions) {
+           throw new InvalidBootstrapURLException("Exceptions during push. Invalid bootstrap url. Please check your "
                                                    + "cluster bootstrap URL");
+
+        }*/
+        ArrayList<Node>failedNodes = new ArrayList<Node>();
+        for (final Node node: cluster.getNodes()){
+            Future<String> val = fetchDirs.get(node.getId());
+            try{
+                String response = val.get();
+                fetchResponseMap.put(node, new Response(response));
+            }catch(Exception e){
+                if (e.getCause() instanceof InvalidBootstrapURLException){
+                    throw new InvalidBootstrapURLException("Not able to find store (" + storeName +
+                            ") in this cluster according to the push URL. BnP job is not able to create new stores now." +
+                            "Please create new stores through Nuage if you think this is the correct cluster you want to push." );
+                }
+                else{
+                    fetchErrors = true;
+                    fetchResponseMap.put(node, new Response(e));
+                    failedNodes.add(node);
+                }
+            }
         }
 
         if(fetchErrors) {
@@ -272,7 +297,7 @@ public class AdminStoreSwapper {
                 }
             }
 
-            if (swapIsPossible) {
+            /*if (swapIsPossible) {
                 // We're good... We'll return the fetchResponseMap.
             } else {
                 String errorMessage = "";
@@ -282,6 +307,10 @@ public class AdminStoreSwapper {
                     errorMessage = "Exception during push. Swap will be aborted.";
                 }
                 throw new VoldemortException(errorMessage);
+
+            }*/
+            if (!swapIsPossible){
+                throw new VoldemortException("Exception during psh. Swap will be aborted", fetchResponseMap.get(failedNodes.get(0)).getException());
             }
         }
         return fetchResponseMap;
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -368,12 +368,13 @@ public class HdfsFetcher implements FileFetcher {
             if(fs != null) {
                 try {
                     fs.close();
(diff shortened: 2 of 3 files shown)


src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -223,10 +224,14 @@ private String fetch(String storeDir) {
             } catch(ExecutionException e) {
                 fetchErrors = true;
                 if(e.getCause() instanceof QuotaExceededException) {
-                    numQuotaExceededException++;
+                    //numQuotaExceededException++;
+                    QuotaExceededExceptions.add(node.getId());
                     fetchResponseMap.put(node, new Response((QuotaExceededException) e.getCause()));
                 } else if(e.getCause() instanceof InvalidBootstrapURLException) {
```
arunthirupathi: When a store is failed to be created, shoudln't it return a meaningful exception instead of sending InvalidBootstrapUrl exceptions ?

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -223,10 +224,14 @@ private String fetch(String storeDir) {
             } catch(ExecutionException e) {
                 fetchErrors = true;
                 if(e.getCause() instanceof QuotaExceededException) {
-                    numQuotaExceededException++;
+                    //numQuotaExceededException++;
+                    QuotaExceededExceptions.add(node.getId());
```
arunthirupathi: Can you please remove the commented code ? If required, we can use git history to recover the code. Having commented code is a bad smell.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1131,12 +1135,18 @@ public void operate() {
 
                             String destinationDir = store.getStoreDirPath() + File.separator + "version-"
                                             + Long.toString(pushVersion);
+
+                            store.setCurrentlyFetching(true);
```
arunthirupathi: This is prone to thread synchronization problems. There are also more corner cases where the job is waiting to be fetched in the queue, then concurrent pushes can happen as well.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -272,7 +297,7 @@ private String fetch(String storeDir) {
                 }
             }
 
-            if (swapIsPossible) {
+            /*if (swapIsPossible) {
```
arunthirupathi: Please remove commented code.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -282,6 +307,10 @@ private String fetch(String storeDir) {
                     errorMessage = "Exception during push. Swap will be aborted.";
                 }
                 throw new VoldemortException(errorMessage);
+
+            }*/
+            if (!swapIsPossible){
+                throw new VoldemortException("Exception during psh. Swap will be aborted", fetchResponseMap.get(failedNodes.get(0)).getException());
```
arunthirupathi: It should be push, u is missing.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -496,8 +496,8 @@ public StreamRequestHandler handleRequest(final DataInputStream inputStream,
             for(String storeName: storeNames) {
 
                 ReadOnlyStorageEngine store = getReadOnlyStorageEngine(metadataStore,
-                                                                       storeRepository,
```
arunthirupathi: Though it is OK for now, you should have an editor setting format only editted lines. It makes it harder to review on big changes.

SidW: Hi Arun,

Thanks for all your comments and sorry for my bad coding style. I also
found out I made mistake commits last time to my own github (that was for
concurrency push issues, haven't done yet) which made my commit timeline
mess up. I have re-forked Voldemort repo and will only commit codes for
invokeFetch() swallows and rethrow exception issue this time.

arunthirupathi:
Sidian, I apologize if I made you feel that your coding style is bad. You are doing a good job, but there are few things that can be improved. I made lot more mistakes than you when I started coding :) . Let us talk more about the parallel pushes issue in person. It is little complicated and I did not have a complete understanding until 2 days back, when I investigated an issue. Now I know little more and I will convey my understanding.

README.md
```diff
@@ -58,6 +58,7 @@ get "k1"
 
 You can find more commands by running```help```
 
+Want to dig into the detailed implementation or even contribute to Voldemort? [A quick git guild for people who want to make contributions to Voldemort](https://github.com/voldemort/voldemort/wiki/A-quick-git-guide-for-people-who-want-to-make-contributions-to-Voldemort).
```
FelixGV: Great.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
FelixGV: +1, let's make sure the check and the setting of this boolean flag is properly synchronized.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1162,6 +1172,9 @@ public void operate() {
                         } catch(Exception e) {
                             throw new VoldemortException("Exception in Fetcher = " + e.getMessage(),
                                                          e);
+                        } finally{
+                            if (store.isCurrentlyFetching())
```
FelixGV: Very good thinking to put this in the finally clause. However, I think you should remove the other ```store.setCurrentlyFetching(false)``` from line 1148, just above, since it makes it possible that another thread would set the flag to true between 1148 and 1177 and then the flag would be clobbered back to false in the finally clause.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -72,6 +73,7 @@
     private volatile ChunkedFileSet fileSet;
     private volatile boolean isOpen;
     private long lastSwapped;
+    private AtomicBoolean currently_fetching;
```
FelixGV: Let's stick with camel case (i.e.: camelCase, not underscore_separation) so that we're more inline with Java coding style conventions.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
FelixGV: Since you use an ```AtomicBoolean```, you might want to look into doing a compare-and-swap operation.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
FelixGV: +1

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -238,9 +243,29 @@ private String fetch(String storeDir) {
...
+        for (final Node node: cluster.getNodes()){
+            Future<String> val = fetchDirs.get(node.getId());
+            try{
+                String response = val.get();
+                fetchResponseMap.put(node, new Response(response));
+            }catch(Exception e){
+                if (e.getCause() instanceof InvalidBootstrapURLException){
+                    throw new InvalidBootstrapURLException("Not able to find store (" + storeName +
+                            ") in this cluster according to the push URL. BnP job is not able to create new stores now." +
+                            "Please create new stores through Nuage if you think this is the correct cluster you want to push." );
```
FelixGV: Hmm... let's not mention Nuage in our error messages, since that system is not open-source (at least not yet) so it doesn't belong here.

Also, can you include ```e``` as a nested exception? This makes it easier to understand the stacktrace and root cause of issues.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
FelixGV: Can we mention which node gave us this exception?

For example: You could add this to the end of the error message string: ```"Exception during push. Swap will be aborted. Re-throwing exception from " + failedNodes.get(0).briefToString()```.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -238,9 +243,29 @@ private String fetch(String storeDir) {
...
+                }
```
FelixGV: Nitpick: let's collapse this into a single line: ```} else {``` with a single space before and after the else.

arunthirupathi:
Closing, this as it is superceeded by a new pull request.

PR #378, opened 2016-02-27
Report metrics for Scheduler and Async Service

arunthirupathi:
Scheduler Service reports tasks currently running
and number of tasks in the queue

Async Service reports cumulative Wait time and
number of tasks waiting in the queue.

### DIFF
--- a/src/java/voldemort/common/service/SchedulerService.java
+++ b/src/java/voldemort/common/service/SchedulerService.java
@@ -252,4 +252,15 @@ public class SchedulerService extends AbstractService {
         }
     }
 
+    @JmxGetter(name = "numActiveTasks", description = "Returns number of tasks executing currently")
+    public long getActiveTasksCount() {
+        return this.scheduler.getActiveCount();
+    }
+
+    @JmxGetter(name = "numQueuedTasks",
+            description = "Returns number of tasks queued for execution")
+    public long getQueuedTasksCount() {
+        return this.scheduler.getQueue().size();
+    }
+
 }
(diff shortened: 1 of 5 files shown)


src/java/voldemort/common/service/SchedulerService.java
```diff
@@ -229,18 +229,22 @@ public SchedulerThreadPool(int numThreads, final String schedulerName) {
                 private AtomicInteger threadCount = new AtomicInteger(0);
 
                 /**
-                 * This function is overridden in order to activate the daemon mode as well as
-                 * to give a human readable name to threads used by the {@link SchedulerService}.
-                 * 
-                 * Previously, this function would set the thread's name to the value of the passed-in
-                 * {@link Runnable}'s class name, but this is useless since it always ends up being a
```
FelixGV: Are the whitespace changes intentional?

arunthirupathi: Nope my eclipse is messing it up. Though I have only format changed line settings.

FelixGV:
The change LGTM.

I guess your intent is for this metric to catch long wait times, not transient waits, right? Because if our reporting infrastructure pokes the metric once per minute, then threads that waited for less than a minute may slip in unnoticed. I think that's fine, as long as we're aware of it when looking at the dashboards.

src/java/voldemort/common/service/SchedulerService.java
arunthirupathi: Fixed it again.

FelixGV: Thanks, LGTM (:

PR #379, opened 2016-02-28
Stop swallowing exception's original message and clarify top-level message

SidW:
AdminStoreSwapper#invokeFetch() used to take charge of collecting exceptions among sub-threads. It swallowed original exceptions and made customers confused when debugging. Now it maintains all sub-thread exception in an list and throw the first exception in the list.

Also modify the InvalidBootstrapurLException message to make it more clear. It is now:
"Not able to find store (" + storeName +") in this cluster according to the push URL. BnP job is not able to create new stores now." + "Please create new stores through Nuage if you think this is the correct cluster you want to push."


FelixGV:
I left a few minor comments in the other PR.

About your commit message:

- I think your commit message got clipped (it says "top" where you probably meant "Stop"?).
- Also, I think we may want the first line of the commit message to be a bit more general so that we can better understand what the concrete change is. Maybe something along the lines of: "Better error message when pushing to a store with storage quota 0".

SidW:
Thanks Felix, have it modified.

FelixGV:
LGTM! I think we should merge this in. Thanks a lot @SidW.

arunthirupathi:
Sidian, let me know if you are comfortable merging this and unit test passes, I will merge this in. Probably after one or two more PRs, we can make you a contributor as well.

SidW:
Hi Arun, I tested my code again and I think it is the time to merge it. Thanks.

arunthirupathi:
The commit Id got messed up, but I was able to merge this into master.

https://github.com/voldemort/voldemort/commit/17f0def9bfd8b8ffdbce9d01cbe339de02e75677

PR #380, opened 2016-03-02
HdfsFetcher's AdminClient is now configurable.

FelixGV:
This AdminClient is used to retrieve quota information from within
the HdfsFetcher. It is a bit messy and we probably should not be
using an AdminClient at all for this, but for now this is just a
quick fix to ensure that we can actually configure this server-side
AdminClient with parameters from VoldemortConfig.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -195,9 +195,10 @@ public class HdfsFetcher implements FileFetcher {
                       String storeName,
                       long pushVersion,
                       MetadataStore metadataStore) throws Exception {
+        // FIXME: Stop using an admin client to talk to yourself, silly Voldemort!
         AdminClient adminClient = null;
         try {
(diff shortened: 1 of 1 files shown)


arunthirupathi:
Felix, What did you guys discuss ? Is this an urgent problem, if so I can crank up the right fix tonight. If not, I would prefer Sidian to take a look at this when he has time.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -195,9 +195,10 @@ public File fetch(String sourceFileUrl,
                       String storeName,
                       long pushVersion,
                       MetadataStore metadataStore) throws Exception {
+        // FIXME: Stop using an admin client to talk to yourself, silly Voldemort!
```
SidW: Well, as Voldemort split his soul into horcruxes. It's possible that he likes to talk with himself :)

SidW:
Hi Arun, I am going to look at it as soon as concurrency push is done!

arunthirupathi:
@SidW  fixed it to not use the AdminClient. So closing this pull request.

PR #381, opened 2016-03-04
RO BnP Bug fixes

arunthirupathi:
1) When BnP HA is activated, try to leave the fetch enabled to
shorten the recovery time.

2) When Version dir does not exist, try to create it before
failing with PersistenceFailureException.

### DIFF
--- a/src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
+++ b/src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
@@ -1943,10 +1943,25 @@ public class AdminServiceRequestHandler implements RequestHandler {
         }
 
         if (response.getDisableSuccess()) {
-            // Then we also want to put the server in offline mode
+            // Voldemort Client can detect and handle failure only at a server
+            // level. Disabling a store inside Server is not possible . Client
+            // creates connection to the server and shares the same connection
+            // for multiple stores. Client failure detection is based on this
+            // connection error. So entire server will be marked as offline for
+            // now.
+
+            boolean roFetchState = metadataStore.getReadOnlyFetchEnabledUnlocked();
+
             VAdminProto.SetOfflineStateRequest offlineStateRequest =
                     VAdminProto.SetOfflineStateRequest.newBuilder().setOfflineMode(true).build();
             handleSetOfflineState(offlineStateRequest);
+
+            if(roFetchState) {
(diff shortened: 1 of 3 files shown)


src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1943,10 +1943,25 @@ private Message handleDisableStoreVersion(VAdminProto.DisableStoreVersionRequest
...
+            // creates connection to the server and shares the same connection
+            // for multiple stores. Client failure detection is based on this
+            // connection error. So entire server will be marked as offline for
+            // now.
+
+            boolean roFetchState = metadataStore.getReadOnlyFetchEnabledUnlocked();
+
             VAdminProto.SetOfflineStateRequest offlineStateRequest =
                     VAdminProto.SetOfflineStateRequest.newBuilder().setOfflineMode(true).build();
             handleSetOfflineState(offlineStateRequest);
```
FelixGV: Should we change the behavior of ```handleSetOfflineState``` instead of leaving it as is, but then re-enabling the fetch right after? Toggling disabled/enabled seems a bit convoluted and hard to reason about.

arunthirupathi: There is no convenient method for just putting the state in offline. 

handleSetOffline needs to pass it all the way down (VoldemortServer to MetadataStore) and add new parameters for setting the individual states.This will make the fix little more complicated. I would like to leave it in this state for now.

FelixGV: What I'm saying is: let's make the offline state consistent: always allow fetches, by modifying the logic in ```MetadataStore.setOfflineState``` so that we don't need to pass parameters through. This makes it easier to reason about what's going on, since by default the offline state means just one thing (unless manually overridden by the operator).

arunthirupathi: I believe, that will work, but I need to run it through the SREs to see where all they use this now.

Some times when they have to take out a node due to degraded SSDs, it would mean running two commands instead of one. 

If Possible, I would like to leave this as it is for now.

arunthirupathi:
After discussing with operations people, changing the way offline state behaves for one case is going to be confusing and might cause surprise to operations. So aborting the fix in its current form.

PR #382, opened 2016-03-05
Resolved concurrency push conflict, refactor quirky AdmintClient used in HdfsFetcher, add hyperlink in README.md

SidW:
1. It was very likely to cause content conflict when multiply jobs try to push to
the same store simultaneously. We add a hashmap in AdminRequestHandler
to check whether a store is currently fetching or not. And if so, we
block the rest of fetching request and throw an exception.

2. The HdfsFetcher used to create an admintClient instance and asked
itself "diskQuotaSizeInKB". This was a quirky used adminClient. We
removed it and passed quota size directly from AdminRequestHandler.

3. Add a hyperlink in README.md to "A quick git guild for people who
want to make contributions to Voldemort".

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
@@ -167,7 +167,7 @@ public class HdfsFetcher implements FileFetcher {
      *
      * FIXME: Refactor test code with dependency injection or scope restrictions so this function is not public.
      *
-     * @deprecated Do not use for production code, use {@link #fetch(String, String, voldemort.server.protocol.admin.AsyncOperationStatus, String, long, voldemort.store.metadata.MetadataStore)} instead.
+     * @deprecated Do not use for production code, use {@link #fetch(String, String, voldemort.server.protocol.admin.AsyncOperationStatus, String, long, voldemort.store.metadata.MetadataStore, Long diskQuotaSizeInKB)} instead.
      */
     @Deprecated
     @Override
@@ -180,7 +180,7 @@ public class HdfsFetcher implements FileFetcher {
      *
      * FIXME: Refactor test code with dependency injection or scope restrictions so this function is not public.
      *
-     * @deprecated Do not use for production code, use {@link #fetch(String, String, voldemort.server.protocol.admin.AsyncOperationStatus, String, long, voldemort.store.metadata.MetadataStore)} instead.
+     * @deprecated Do not use for production code, use {@link #fetch(String, String, voldemort.server.protocol.admin.AsyncOperationStatus, String, long, voldemort.store.metadata.MetadataStore, Long diskQuotaSizeInKB)} instead.
      */
     @Deprecated
     @Override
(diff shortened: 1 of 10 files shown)


src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1085,14 +1093,35 @@ private String swapStore(String storeName, String directory) throws VoldemortExc
...
-            asyncService.submitOperation(requestId, operation);
+                                                                                    pushVersion,
+                                                                                    diskQuotaSizeInKB);
+            AdminServiceRequestHandler.storeLock.lock();
+
+            try {
+                if (AdminServiceRequestHandler.fetchingStores.containsKey(storeName)) {
+                    int previousRequestID = AdminServiceRequestHandler.fetchingStores.get(storeName);
+                    if (!asyncService.isComplete(previousRequestID, false))
+                        throw new VoldemortException("The store: " + storeName + " is currently block since it is fetching data");
```
FelixGV: Typo: block => blocked

Since you have the request ID that is currently in progress, I suggest including it in the exception message. That would be very useful information, I think.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1085,14 +1093,35 @@ private String swapStore(String storeName, String directory) throws VoldemortExc
...
                                                                                     storeName,
                                                                                     fetchUrl,
-                                                                                    pushVersion);
-            asyncService.submitOperation(requestId, operation);
+                                                                                    pushVersion,
+                                                                                    diskQuotaSizeInKB);
+            AdminServiceRequestHandler.storeLock.lock();
+
+            try {
+                if (AdminServiceRequestHandler.fetchingStores.containsKey(storeName)) {
```
FelixGV: It would probably be safer if this whole if block was inside of a synchronize block.

src/java/voldemort/server/protocol/admin/AsyncOperationService.java
```diff
@@ -81,21 +81,28 @@ public synchronized void submitOperation(int requestId, AsyncOperation operation
      * @param requestId Id of the request
      * @return True if request is complete, false otherwise
      */
-    public synchronized boolean isComplete(int requestId) {
-        if(!operations.containsKey(requestId))
-            throw new VoldemortException("No operation with id " + requestId + " found");
+    public synchronized boolean isComplete(int requestId, boolean remove) {
+        //if requestId not found, return true
+        if (!operations.containsKey(requestId))
```
FelixGV: I don't think it's a good idea to change this behavior. Instead, for the use case where you consider a non-existent requestId to be equivalent to a completed one, you should surround the call in a try catch and ignore the exception.

FelixGV:
Thanks a lot! These are some great changes. I left a few minor comments. LGTM otherwise.

src/java/voldemort/server/protocol/admin/AsyncOperationService.java
arunthirupathi: There are only 1 other caller to this method and the caller is modified too.

arunthirupathi: I take it back, the caller is not modified. The behavior needs to be preserved in the handleAsyncStatus

Felix, the problem is readonly fetch calls isComplete which has side effects. If fetch read only removes it pre-maturely, then handleAsyncStatus will throw.

FelixGV: Ok well... It still seems a bit unintuitive that any arbitrary requestId that may or many not have ever happened is considered completed. In any case, if we leave it like that, we should at least update the JavaDoc accordingly.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -167,7 +167,7 @@ private HdfsFetcher(VoldemortConfig config,
      *
      * FIXME: Refactor test code with dependency injection or scope restrictions so this function is not public.
```
FelixGV: Thanks a lot for this commit @SidW (:

Now that we're properly passing in the quota as a parameter, rather than getting it from the ```MetadataStore``` instance, I believe we can finally get rid of these two bogus test-only functions.

Do you think you could remove them and refactor the impacted tests accordingly?

-F

arunthirupathi:
Thanks @FelixGV . @SidW  the context here is I did some refactorings for a future change, which never materialized. So you can remove the additional parameters if required.

The additional parameters are cause for the bad quota code, as it made the buggy code very easy to write, instead of passing down the parameter.

PR #387, opened 2016-03-18
Support for expanding store and view names

tempredirect:
General idea is to make the stores.xml less boiler plate by allowing
identical store definitions to to be expressed as a single `<store>`
element with `<name>` tag in the form `prefix-{name1,name2}`. This will
result in two stores being defined `prefix-name1` and `prefix-name2`
each with the same store definition.

Views get a similar treatment but with a regex matching against store
names and expanding the name field with the match so

```
    <view>
        <name>prefix-$1-view</name>
        <view-of>prefix-(.*)</view-of>
        <view-of-is-regexp>true</view-of-is-regexp>
```

Will match the above stores `prefix-name1` and `prefix-name2` and
generate views `prefix-name1-view` and `prefix-name2-view` backed by the
corresponding store.


arunthirupathi:
@tempredirect What is the advantage of implementing this feature on the Server, versus implement this as a script, which can do a similar thing.

Using regex and others looks good on the surface, but there is no way to review it on the server side and if it aborts in the middle, there is no way to roll back.

would you be willing to modify this on the client side to generate the modified stores.xml and reviewing it and do not modify the server side APIs ?

Please let me know if you need further information.

arunthirupathi:
@tempredirect  please let me know if you have a different opinion. I would love to hear your thoughts on the issue. 

If there are operations people who runs voldemort, you guys can chime in here too. But for me the non predictability of the command and this happening on the server side is little concerning. I would expect the regex expansion to be done by a tool or by the client side code and then add one store at a time to the server.

tempredirect:
Sorry, was off yesterday.. have been catching up.

I'm completely ok with you closing the request, I won't take offence.

First a little confession/background, this isn't strictly my work, it is actually quite old the original patch internally is ~4 years old. My employer has given me permission to release it, more as a tryout on actually releasing stuff. 

I think that original motivation has more to do with our internal scm and build system than any great master plan. At the point this patch was written stores.xml would be deployed directly (almost) from the scm without the opportunity hook in any sort of build/cmdline tool.

Building a cmdline tool, doesn't really appeal right at the moment as it would either have to have it's own stores.xml parser (that understood the new elements) or the internal api's for parsing would exist with a kinda offline only mode. It might happen, but we're likely to keep taking the pain of rebasing this patch onto the open source tree for the time being.

Hopefully there might be a couple more patches in the pipeline.

arunthirupathi:
I see never mind, thanks for the patch, sorry that I won't merge this in.

arunthirupathi:
@tempredirect  on reading it again, I thought there are few important changes to Voldemort.

4 years back, Voldemort only way to add/remove stores is get the stores.xml add or remove your stores and post it back. couple of years back Voldemort added the feature to add only one store.

In newer versions instead of storing all the stores.xml in a single file, it stores one store in a file under the STORES directory.

PR #388, opened 2016-03-22
Check server state when bring nodes back

SidW:
Fixed the issue when starting up Voldemort in offilne state, it actually
goes online and keeps listening client requests.

### DIFF
--- a/src/java/voldemort/server/VoldemortServer.java
+++ b/src/java/voldemort/server/VoldemortServer.java
@@ -382,7 +382,13 @@ public class VoldemortServer extends AbstractService {
         JNAUtils.tryMlockall();
         logger.info("Starting " + basicServices.size() + " services.");
         long start = System.currentTimeMillis();
-        boolean goOnline = true;
+
+        boolean goOnline;
+        if (getMetadataStore().getServerStateUnlocked() == MetadataStore.VoldemortState.OFFLINE_SERVER)
+            goOnline = false;
+        else
+            goOnline = true;
+
         for(VoldemortService service: basicServices) {
             try {
                 service.start();
@@ -391,6 +397,7 @@ public class VoldemortServer extends AbstractService {
                 goOnline = false;
             }
         }
+
         if (goOnline) {
             startOnlineServices();
         } else {
--- a/test/unit/voldemort/client/OfflineStateTest.java
+++ b/test/unit/voldemort/client/OfflineStateTest.java
@@ -20,9 +20,15 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.ConnectException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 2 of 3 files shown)


src/java/voldemort/server/VoldemortServer.java
```diff
@@ -382,7 +382,15 @@ protected void startInner() throws VoldemortException {
...
         long start = System.currentTimeMillis();
-        boolean goOnline = true;
+
+        boolean goOnline;
+        if (getMetadataStore().getServerStateUnlocked() == MetadataStore.VoldemortState.NORMAL_SERVER)
+            goOnline = true;
+        else if (getMetadataStore().getServerStateUnlocked() == MetadataStore.VoldemortState.OFFLINE_SERVER)
+            goOnline = false;
+        else
+            throw new VoldemortException("Invalid Server state. Neither online nor offline found");
```
arunthirupathi: The server has 3 states, online, offline and rebalancing. This makes the server not to start when that is in the Rebalancing mode. I would like to have only 2 cases here, offline and everything else.

test/unit/voldemort/client/OfflineStateTest.java
```diff
@@ -262,4 +268,23 @@ public void testStateTransitions() throws InterruptedException {
...
+
+        String baseDirPath = server.getVoldemortConfig().getVoldemortHome();
+        String parentDirPath = Paths.get(baseDirPath).getParent().toString();
+        List<StoreDefinition> storeDefs = server.getMetadataStore().getStoreDefList();
+        try {
+            VoldemortConfig config =
+                    ServerTestUtils.createServerConfigWithDefs(useNio, 0, parentDirPath, cluster, storeDefs, new Properties());
+            VoldemortServer newServer = ServerTestUtils.startVoldemortServer(socketStoreFactory, config, cluster);
+            fail("should have thrown an connection exception");
+        }catch (VoldemortException e){
```
arunthirupathi: I would just handle Catch UnreachableStoreException and also check that the inner exception is ConnectException.class.

test/unit/voldemort/client/OfflineStateTest.java
```diff
@@ -262,4 +268,23 @@ public void testStateTransitions() throws InterruptedException {
...
+            assertEquals(UnreachableStoreException.class, e.getClass());
+        }
+    }
```
arunthirupathi: I would also like to add that the server that was brought up in offline mode can go to online mode. by calling the appropriate methods.

test/unit/voldemort/client/OfflineStateTest.java
```diff
@@ -262,4 +268,23 @@ public void testStateTransitions() throws InterruptedException {
         assertTrue(testOnlineTraffic());
         assertTrue(testSlopStreaming());
     }
+
+    @Test
```
arunthirupathi: I would also like to add one more test, where a server stopped in online mode, can be brought up in offline mode. The steps that are done in the test, need to be documented and given to SREs.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -386,10 +386,8 @@ protected void startInner() throws VoldemortException {
         boolean goOnline;
         if (getMetadataStore().getServerStateUnlocked() == MetadataStore.VoldemortState.NORMAL_SERVER)
```
arunthirupathi: Rebalance Server should still come as online. This would make the rebalance server to go offline.

arunthirupathi:
The test looks good now. If you can make minor tweak to the original logic and squash your commit, I will merge the commit.

SidW:
Thanks Arun! I should have been more careful. I have modified that and
rebased my commits. Ready to be merged😀

arunthirupathi:
Rebased and merged  https://github.com/voldemort/voldemort/commit/af3d605f9b953f1caa0bd36cde3f1e3a0635da3e

PR #389, opened 2016-03-22
Hdfs FileSystem handles are leaked in Fetch

arunthirupathi:
Issue : Hdfs FileSystem has a cache, which caches all the file system
returned by Scheme ( swebhdfs) + authority ( namenode) + user ( user
running the get filesystem, voldemort account ). While closing the file
system, the close should happen from the same user context. But in
Voldemort we are closing as app user, this causes the filesystem handle
to be leaked.

Previously in voldemort webhdfs file system handles were leaked.
Apparently webhdfs file system handles are very cheap. But in SwebHdfs
they have the security certificate embedded in them. This causes them to
be very big. 1748 objects account for 3 GB and this killed the server.

Heap Dump analysis
WebHdfsFileSystem - 3768 Objects - 80 MB
SWebHdfsFileSystem - 1748 Objects - 3 GB

Solution : Disable the caching, as we re-login for every filesystem
creation and the caching is not intended for the FileSystem handles
anyway.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/HadoopUtils.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/HadoopUtils.java
@@ -33,7 +33,12 @@ import java.util.TreeMap;
 import java.util.regex.Pattern;
 
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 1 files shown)


arunthirupathi:
Succeeded by the pull request
https://github.com/voldemort/voldemort/pull/390

PR #390, opened 2016-03-22
Hdfs FileSystem handles are leaked in Fetch

arunthirupathi:
Issue : After the HadoopFileSystem object is created, the validity of
the fileSystem is verified by doing a sample operation. If the operation
fails the Hadoop FileSystem object is leaked. This object should be
cleaned up by the Garbage collection, but all the FileSystem objects
are cached, so this is leaked. When voldemort server is used with
secure webhdfs (swebhdfs) file system it leaks enough memory to kill
the servers eventually.

Previously in voldemort webhdfs file system handles were leaked.
Apparently webhdfs file system handles are very cheap. But in SwebHdfs
they have the security certificate embedded in them. This causes them to
be very big.

Heap Dump analysis
WebHdfsFileSystem - 3768 Objects - 80 MB
SWebHdfsFileSystem - 1748 Objects - 3 GB

Solution :
1)  Disable the caching, as we re-login for every filesystem
creation and the caching is not intended for the FileSystem handles
anyway. Re-visit this when kerberos caching is enabled as default.
2) Clean up the file handles on the error cases. Traced down all the
handles and cleaned them up on the error path.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/HadoopUtils.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/HadoopUtils.java
@@ -32,8 +32,14 @@ import java.util.Properties;
 import java.util.TreeMap;
 import java.util.regex.Pattern;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.*;
+import org.apache.hadoop.fs.CommonConfigurationKeys;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+import org.apache.hadoop.hdfs.web.HftpFileSystem;
 import org.apache.hadoop.io.SequenceFile;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.FileInputFormat;
@@ -47,8 +53,8 @@ import voldemort.serialization.json.JsonTypeDefinition;
 import voldemort.server.VoldemortConfig;
 import voldemort.store.readonly.fetcher.ConfigurableSocketFactory;
 import voldemort.utils.ExceptionUtils;
-import voldemort.utils.UndefinedPropertyException;
 import voldemort.utils.Props;
+import voldemort.utils.UndefinedPropertyException;
 
 /**
  * Helper functions for Hadoop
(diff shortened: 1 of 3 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/HadoopUtils.java
```diff
@@ -326,10 +334,34 @@ private static Configuration getConfiguration(VoldemortConfig voldemortConfig, S
         hadoopConfig.set("hadoop.security.group.mapping",
                    "org.apache.hadoop.security.ShellBasedUnixGroupsMapping");
 
+        // sourceScheme is null for file based paths, used in testing.
+        if(sourceScheme != null) {
+            /*
+             * Hadoop FileSystem class caches the FileSystem objects based on
+             * the scheme , authority and UserGroupInformation.
+             * 
+             * The default config was to generate new UserGroupInformation for
```
singhsiddharth: Do we use the default config in prod ? Can you confirm please.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
```diff
@@ -108,7 +110,12 @@
     public HdfsFailedFetchLock(VoldemortConfig config, Props props) throws Exception {
         super(config, props);
         fileSystem = HadoopUtils.getHadoopFileSystem(config, clusterDir);
-        initDirs();
+        try {
+            initDirs();
+        } catch(Exception ex) {
```
squarY: Shall we catch the Throwable instead of Exception here? So that we could catch all types of error/exception.

arunthirupathi: I don't know the difference between both of them. I thought Exception is basic for all java exceptions.

arunthirupathi:
Rebased and  Merge with the commit https://github.com/voldemort/voldemort/commit/66bfe3b85f810510210210d4095535c18dd14517

gaojieliu:
Hi @arunthirupathi ,
I have a quick question:
Since the filesystem cache is using scheme + authority + user group info as the cache key, why do we have so many cache objects in prod? It is caused by the fact we are using a lot of clusters in prod?

arunthirupathi:
@gaojieliu  that is a good question. UserGroupInfo represents a login for the user and not the actual user itself. So every time your login get a new UserGroupInfo object. 

Voldemort default config is to login every time ( we tried login only every 30 minutes, but we ran into lots of file handle issue). Now on looking at the cache, I understand why we were running into those issues.  At the end of the every fetch, the file handle is closed so if it is in use by other fetches, it randomly fails others.

gaojieliu:
@arunthirupathi ,
Thanks a lot for the explanations.
Since we will login every time, the cache will be useless since we will generate a new login object and the object id will be different.
In the code, we also close the file handle after every fetch.
Based on the above facts, other fetches should not reuse the previous file handle, right? since every fetch will use a new login and  it will generate a new file handle every time, right?

Not sure about this statement: 
if it is in use by other fetches, it randomly fails others.

gaojieliu:
Got it, this file handle issue will happen if we reuse the user token.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/HadoopUtils.java
FelixGV: Yes we do. I tried implementing the re-use of UGI instances in the past, but it was unstable and I didn't have time to fully debug it, so I made it optional and hid it behind a config. The config's JavaDoc does have a warning that the feature is unstable.

arunthirupathi: Felix, now i understand this area completely, I can explain what is going on.

PR #391, opened 2016-03-24
Update version after updating the data

arunthirupathi:
Currently the version is updated before the data and it
makes the client to read the wrong values.

Update the Version after updating the data, so when the client
reads they have the correct value.


src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -1419,16 +1424,7 @@ public void updateRemoteMetadataPair(List<Integer> remoteNodeIds,
...
-                updateRemoteMetadataPair(currentNodeId,
-                                         clusterKey,
-                                         clusterValue,
-                                         storesKey,
-                                         storesValue);
-            }
+
         }
 
         /**
```
SidW: Looks good for me :)

PR #394, opened 2016-03-28
Expose data points in FileFetcher and create autometrics sensors for them

gaojieliu:
This change is mostly to expose more aggregated metrics for HDFS data pushes:
1. totalBytesFetched : the total bytes transferred from HDFS so far;
2. totalFetchRetries : the total fetch retry number so far;
3. totalCheckSumFailures : the total data file checksum failures
happened so far;
4. totalAuthenticationFailures : the total authentication failures
happened so far;
5. totalFileNotFoundFailures : the total file-not-found failures
happened so far;
6. totalFileReadFailures : the total HDFS file read failures happened so
far;
7. totalQuotaExceedFailures : the total quota exceed failures happened
so far;
8. totalUnauthorizedStoreFailures : the total unauthorized store push
failures happened so far;
9. parallelFetches : the total number of active fetches right now;
10. totalFetches : the total HDFS fetch number so far;
11. totalIncompleteFetches : the total incomplete fetch number so far;
12. totalDataFetchRate : the total data fetch rate right now;

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
@@ -2,6 +2,7 @@ package voldemort.store.readonly.fetcher;
 
 import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -11,10 +12,15 @@ import java.util.Comparator;
 import java.util.Date;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.security.authentication.client.AuthenticationException;
 import org.apache.log4j.Logger;
 
+import org.apache.log4j.spi.ThrowableInformation;
 import voldemort.annotations.jmx.JmxGetter;
+import voldemort.store.quota.QuotaExceededException;
+import voldemort.store.readonly.UnauthorizedStoreException;
 import voldemort.utils.ByteUtils;
+import voldemort.utils.ExceptionUtils;
 import voldemort.utils.Time;
 
 /**
@@ -36,6 +42,10 @@ public class HdfsCopyStats {
     public static final String STATS_DIRECTORY = ".stats";
     private final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
     private static final Logger logger = Logger.getLogger(HdfsCopyStats.class);
+
+    // Store name for the data pull
+    private String storeName = null;
+
 
     private static void deleteExtraStatsFiles(File statsDirectory, int maxStatsFile) {
         if(maxStatsFile <= 0) {
@@ -143,14 +153,18 @@ public class HdfsCopyStats {
                          boolean enableStatsFile,
                          int maxVersionsStatsFile,
                          boolean isFileCopy,
-                         HdfsPathInfo pathInfo) {
+                         HdfsPathInfo pathInfo,
+                         String storeName) {
         this.sourceFile = source;
         this.totalBytesTransferred = 0L;
         this.bytesTransferredSinceLastReport = 0L;
         this.pathInfo = pathInfo;
+        this.storeName = storeName;
         this.lastReportNs = System.nanoTime();
         this.startTimeMS = System.currentTimeMillis();
         initializeStatsFile(destination, enableStatsFile, maxVersionsStatsFile, isFileCopy);
+        // Register the current HdfsCopyStats to the aggregated stats
+        HdfsFetcherAggStats.getStats().addStoreCopyStats(storeName, this);
     }
 
     public void recordBytesWritten(long bytesWritten) {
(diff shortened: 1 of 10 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
```diff
@@ -143,14 +153,18 @@ public HdfsCopyStats(String source,
...
         this.sourceFile = source;
         this.totalBytesTransferred = 0L;
         this.bytesTransferredSinceLastReport = 0L;
         this.pathInfo = pathInfo;
+        this.storeName = storeName;
         this.lastReportNs = System.nanoTime();
         this.startTimeMS = System.currentTimeMillis();
         initializeStatsFile(destination, enableStatsFile, maxVersionsStatsFile, isFileCopy);
+        // Register the current HdfsCopyStats to the aggregated stats
+        HdfsFetcherAggStats.getStats().addStoreCopyStats(storeName, this);
```
arunthirupathi: instead of calling , getStats method on each one of the invocation, I would prefer to have the Static variable here.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcherAggStats.java
```diff
@@ -0,0 +1,190 @@
...
+    @JmxGetter(name = "totalIncompleteFetches", description = "The total incomplete fetch number so far.")
+    public synchronized long getTotalIncompleteFetches() {
+        return totalIncompleteFetches;
+    }
+
+    @JmxGetter(name = "totalDataFetchRate", description = "The total data fetch rate right now.")
+    public synchronized double getTotalDataFetchRate() {
+        double fetchRate = 0;
+        for (HdfsCopyStats copyStats : copyStatsMap.values()) {
+            fetchRate += copyStats.getBytesTransferredPerSecond();
```
arunthirupathi: The fetchRate is not often updated and might be calculated at different intervals. EventThrottler has a rate sensor, we should output that for totalFetchRate.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -1,12 +1,12 @@
 /*
  * Copyright 2008-2013 LinkedIn, Inc
- *
```
arunthirupathi: This file can be reverted, as it only has the WhiteSpace change.

contrib/hadoop-store-builder/test/voldemort/store/readonly/fetcher/HdfsFetcherTest.java
```diff
@@ -168,4 +168,291 @@ public void testFetch() throws Exception {
         checkSumFile.delete();
 
     }
+
+    public void testAggStatsWithValidFile() throws Exception {
+        HdfsFetcherAggStats stats = HdfsFetcherAggStats.getStats();
```
arunthirupathi: I believe we are validating all parameters here. Can we validate only the things, that we expect to change.

Or other way is to have a copyConstructor for the HdfsFetcherAggStats and also make the method implement equals. This will reduce the amount of boiler plate code in the test.

arunthirupathi:
Gaojie, thanks for making the changes. I left few minor comments, but the code looks good. Once you can rebase ( and resolve the merge conflict), Sidian's changes are conflicting with your changes, we can merge this in.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
gaojieliu: Sure, will do.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcherAggStats.java
gaojieliu: Good suggestion!
I only have one concern, since the rate sensor in EventThrottle only works when we setup throttle config in Voldemort server, which means this sensor rate will have nothing if we disable throttle control.
I assume we will always enable throttle control in the real production scenario, so I can reuse this sensor, otherwise I need to create a new sensor to capture the data rate.
@arunthirupathi ,
Is there any user case we will disable throttle control?

src/java/voldemort/server/VoldemortServer.java
gaojieliu: Could I keep this change since I was using Felix's formatting rule?
The unnecessary white spaces at the end of line should be removed.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcherAggStats.java
```diff
@@ -0,0 +1,190 @@
...
+    // Total file not found failures
+    private long totalFileNotFoundFailures;
+    // Total quota exceed failures
+    private long totalQuotaExceedFailures;
+    // Total unauthorized store failures
+    private long totalUnauthorizedStoreFailures;
+
+    // Parallel active fetch number
+    // This metric is only used to capture the active fetches number right now
+    private int parallelFetches;
```
stotch: Rename this to activeFetches to be more general.

parallelFetches can be used later on if we want to control parallelism/priority of fetches and want to track how many are pulling data in parallel.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsCopyStats.java
```diff
@@ -161,6 +175,41 @@ public void recordBytesWritten(long bytesWritten) {
...
+        if (ExceptionUtils.recursiveClassEquals(e, AuthenticationException.class)) {
+            HdfsFetcherAggStats.getStats().authenticateFailed();
+        } else if (e instanceof QuotaExceededException) {
+            HdfsFetcherAggStats.getStats().quotaCheckFailed();
+        } else if (e instanceof UnauthorizedStoreException) {
+            HdfsFetcherAggStats.getStats().unauthorizedStorePush();
+        } else if (e instanceof FileNotFoundException) {
+            HdfsFetcherAggStats.getStats().fileNotFound();
+        } else {
+            HdfsFetcherAggStats.getStats().fileReadFailed();
```
stotch: This should probably tracked as an "unknown" file fetch failure, probably best stored in unknownFileFetchFailures (or similarly named).

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/HdfsFetcher.java
```diff
@@ -348,6 +364,9 @@ private File fetchFromSource(String sourceFileUrl,
             if(stats != null) {
                 stats.complete();
             }
+            if (!isCompleteFetch) {
```
stotch: I think most of our code has a space between the operator and the operand. For consistency, should probably put a space between ! and isCompleteFetch.

PR #395, opened 2016-03-28
update checkout from disk instead of memory buffer.

SidW:
Added second checksum validation after hdfsFetcher had pulled data to local disk. The fetcher now exams checksum values twice:
1. when the data reaches memory buffer.
2. when the data is written in hard disk. 

I didn't remove the memory buffer check since it kept writing checksum data into HdfsCopyStats while data transition.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
@@ -1,7 +1,9 @@
 package voldemort.store.readonly.fetcher;
 
+import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.FilterInputStream;
 import java.io.IOException;
@@ -15,11 +17,12 @@ import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.log4j.Logger;
 
+import voldemort.VoldemortException;
 import voldemort.server.protocol.admin.AsyncOperationStatus;
 import voldemort.server.protocol.admin.AsyncOperationStoppedException;
 import voldemort.store.readonly.checksum.CheckSum;
 import voldemort.store.readonly.checksum.CheckSum.CheckSumType;
-import voldemort.utils.ExceptionUtils;
+import voldemort.utils.ByteUtils;
 
 
 public class BasicFetchStrategy implements FetchStrategy {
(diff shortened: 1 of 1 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
```diff
@@ -208,6 +216,26 @@ public BasicFetchStrategy(HdfsFetcher fetcher,
                 }
             }
         }
+
+        //second time checksum validation. Check if the local file is consistent with the buffer
+        if (bufferCheckSumGenerator != null){
+            CheckSum fileCheckSum = CheckSum.getInstance(checkSumType);
```
arunthirupathi: This should probably be called fileCheckSumGenerator.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
```diff
@@ -208,6 +216,26 @@ public BasicFetchStrategy(HdfsFetcher fetcher,
...
+            BufferedInputStream in = new BufferedInputStream(new FileInputStream(dest));
+            int read;
+
+            try {
+                while ((read = in.read(buffer)) >= 0) {
+                    fileCheckSum.update(buffer, 0, read);
+                }
+
+                if (ByteUtils.compare(checkSum, fileCheckSum.getCheckSum()) != 0)
+                    throw new VoldemortException("Local file: " + dest.getAbsolutePath() +
```
arunthirupathi: Can you also print what is the fileCheckSum and buffer checksum in the error logs ?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/fetcher/BasicFetchStrategy.java
```diff
@@ -1,11 +1,15 @@
...
 
+import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.file.Files;
```
arunthirupathi: Files and Paths can be removed, if they are not used.

arunthirupathi:
looks good, will merge this in.

arunthirupathi:
Rebased and committed. https://github.com/voldemort/voldemort/commit/64726a93da7e8f8a0175f5d073e751caf84c0c33

FelixGV:
@SidW, @arunthirupathi,

I assume this change is to catch issues where the filesystem may fail silently by corrupting the data without reporting an error at write-time?

I have two related questions:

1. Did we measure the impact in terms of total fetch time?
2. What about IO, are we trashing the page cache? Conversely, if we suspect that there is silent data corruption going on in an SSD, is there a possibility that the OS will have a correct (non-corrupt) data file content in its page cache, and thus also silently fail the second check, since what Java thinks it is loading content from the SSD but it may actually be coming from the OS' page cache?

@gnb, do you have any thoughts on the above?

-F

SidW:
1. I haven't measure the impact regarding to the time. 
2. Didn't even think about this issue and would know the truth :)

PR #397, opened 2016-04-01
JsonStoreBuilder throws ClassCastExceptions when store is configured with identity serializers

jalkjaer:
This PR fixes the problem by  base64 decoding the key/value data where identity (byte arrays) are expected.  Works for both key and value serialization

i.e. Store xml defined as 
```xml
..
 <key-serializer>
    <type>string</type>
 </key-serializer>
 <value-serializer>
    <type>identity</type>
 </value-serializer>
..
```
Will now succesfully generate stores with input files like:
```
"key1" "AAE="
"key2" "AAI="
```

### DIFF
--- a/src/java/voldemort/store/readonly/JsonStoreBuilder.java
+++ b/src/java/voldemort/store/readonly/JsonStoreBuilder.java
@@ -36,6 +36,7 @@ import java.util.Map.Entry;
 import java.util.Random;
 import java.util.Set;
 
+import com.google.common.io.BaseEncoding;
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
 
@@ -46,10 +47,7 @@ import voldemort.cluster.Cluster;
 import voldemort.cluster.Node;
 import voldemort.routing.RoutingStrategy;
 import voldemort.routing.RoutingStrategyFactory;
-import voldemort.serialization.DefaultSerializerFactory;
-import voldemort.serialization.Serializer;
-import voldemort.serialization.SerializerDefinition;
-import voldemort.serialization.SerializerFactory;
+import voldemort.serialization.*;
 import voldemort.serialization.json.EndOfFileException;
 import voldemort.serialization.json.JsonReader;
 import voldemort.store.StoreDefinition;
@@ -68,8 +66,8 @@ import com.google.common.collect.Maps;
 
 /**
  * Build a read-only store from given input.
- * 
- * 
+ *
+ *
  */
 public class JsonStoreBuilder {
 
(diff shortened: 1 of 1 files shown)


arunthirupathi:
@jalkjaer  I am very confused. You are using identify serializer and using base64 .  In voldemort context identity means input is byte array and serialized to byte array.  Which in turn means no serialization.

Can you please tell us what you are trying to do ? Why is not possible or why is it sub-optimal ? What is your solution ?

If you want to use byte array, you should probably use Avro serialization and have bytes in there. Base 64 encoding is good for HTTP style protocol where binary characters pose a problem. This is very hacky solution and in this current form, I am going to close it. I would encourage you to start a discussion on the groups, so that community can help you.

jalkjaer:
Let me elaborate:

To populate a ReadOnly store, you normaly run the Hadoop base job. An alternative is to use the cli tool `voldemort.store.readonly.JsonStoreBuilder` which works locally without all the complexities of haddop. This is convenient for generating test stores as you simply the tool to a text file. 
The problem with JsonStoreBuilder is that it cannot generate stores files if the store is defined with the "identity"  key or value serializer. 

I'm not ready to change the store definition as that will require changes to all clients that read from the store as the byte content will be altered due to the change in serialization format.

arunthirupathi:
But isn't this still giving different behaviors for Product and test only code ? If you define the same schema in the Hadoop, it is going to end up with different key/value bytes. This is confusing and it does more bad than good in my opinion. If it solves a specific problem for your use case, you can have it in your fork, but I don't see it good enough to be good for the community.

Please correct me if my understanding is incorrect.

jalkjaer:
Both our prod and test stores are configured with identity value
serialization. Our hadoop output is  sequencefiles of string,
byteswritable. The mapper is then trivial

arunthirupathi:
@jalkjaer  I am really confused. 

JsonStoreBuilder is used only for Tests as you mentioned. I searched the code again and Voldemort except in Rest co-ordinator code path, does not have base64 encoding/decoding.

With the pull request you sent, JsonStoreBuilder will use base64 for encoding/decoding, but prod code path will use some other way. This is inconsistent behavior.

Base64 is used to serialize bytes especially when used with HTTP protocols to work with binary data. But combining Json and base64 is ugly hacking, though it may be easier to use it in shell or some other testing.

Please let me know if something is not clear.

jalkjaer:
It's unclear to me why you are talking about the rest client and prod code. JsonStoreBuilder is a commandline tool used to create ReadOnly store files from a text input file. E.g. its not part of any "production" call path. 

The JsonStoreBuilder in master cannot create store files, if the store definition has
```
 <value-serializer>
    <type>identity</type>
 </value-serializer>
```
It fails with an ClassCast exception as described in my previous comments.

The PR changes how the JsonStoreBuilder interprets the input text file when the store definition has `<type>identity</type>` in either key- or value-serializer. The PR doesn't change how the stores are being generated. Given that JsonStoreBuilder  reads a text file I don't think its a very controversial or hackish to assume Base64 in this case as that is the de facto standard for storing binary content in a text file/stream

PR #398, opened 2016-04-15
The AdminClient's verifyOrAddStore is vulnerable to connectivity issues.

FelixGV:
This commit makes the following changes:
- verifyOrAddStore() is now more resilient to various kinds of exceptions.
- ExceptionUtils.recursiveClassEquals() can now look for many exception types.
- VoldemortBuildAndPushJob now logs verifyOrAddStore's exceptions.

### DIFF
--- a/src/java/voldemort/client/protocol/admin/AdminClient.java
+++ b/src/java/voldemort/client/protocol/admin/AdminClient.java
@@ -25,6 +25,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.StringReader;
 import java.io.UnsupportedEncodingException;
+import java.net.ConnectException;
 import java.net.Socket;
 import java.net.SocketException;
 import java.nio.channels.Channels;
@@ -108,6 +109,7 @@ import voldemort.store.system.SystemStoreConstants;
 import voldemort.store.views.ViewStorageConfiguration;
 import voldemort.utils.ByteArray;
 import voldemort.utils.ByteUtils;
+import voldemort.utils.ExceptionUtils;
 import voldemort.utils.MetadataVersionStoreUtils;
 import voldemort.utils.NetworkClassLoader;
 import voldemort.utils.Pair;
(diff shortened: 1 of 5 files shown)


src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -1875,10 +1877,16 @@ public void verifyOrAddStore(StoreDefinition newStoreDef, String localProcessNam
...
-                    continue;
+                } catch (Exception e) {
+                    if (ExceptionUtils.recursiveClassEquals(e,
+                                                            UnreachableStoreException.class,
+                                                            IOException.class)) {
+                        logger.warn("Failed to contact " + node.briefToString() + " in order to validate the StoreDefinition.");
+                        unreachableNodes.add(node);
+                        continue;
+                    } else {
+                        throw new VoldemortException("Unhandled exception during verifyOrAddStore() !", e);
```
arunthirupathi: Can we just rethrow the e, do reduce the semantic change.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -1875,10 +1877,16 @@ public void verifyOrAddStore(StoreDefinition newStoreDef, String localProcessNam
                 try {
                     // Get all StoreDefinitions from each nodes in the cluster
                     remoteStoreDefs = metadataMgmtOps.getRemoteStoreDefList(nodeId).getValue();
-                } catch (UnreachableStoreException e) {
-                    logger.warn("Failed to contact " + node.briefToString() + " in order to validate the StoreDefinition.");
-                    unreachableNodes.add(node);
-                    continue;
+                } catch (Exception e) {
+                    if (ExceptionUtils.recursiveClassEquals(e,
```
arunthirupathi: Can you leave a comment here saying that?

Admin Client Checkout code may throw, Connect, Socket or EOFException, so checking for IOException as well.

src/java/voldemort/client/protocol/admin/AdminClient.java
FelixGV: This function ```throws UnreachableStoreException, VoldemortException``` so it does not allow throwing the untyped ```Exception e```. Therefore, wrapping whatever ```Exception``` in ```VoldemortException``` is not a semantic change, I think.

src/java/voldemort/client/protocol/admin/AdminClient.java
FelixGV: Done in the latest amend.

src/java/voldemort/client/protocol/admin/AdminClient.java
arunthirupathi: Can you catch the VoldemortException then ? The Call stack shows SocketException was wrappedInside VoldemortException.

FelixGV: Ok, done.

arunthirupathi:
Looks good.

PR #400, opened 2016-04-23
Metadata check to include quota check

arunthirupathi:
1) Reliably set Quota on all nodes.
2) Meta check will check for quota on all nodes.
3) Meta check will ignore 0 length RO files
4) Meta check will skip nodes with 0 partitions when verifying
store can be fetched.
5) ReplaceNodeCLI node validation and ignore all errors from failing
node.
6) Set Quota to report the correct node it is going to run against.
7) When a store is created, set the default quota directly instead
of using an admin client to write to the same node.

All operability improvement for working with quota.

### DIFF
--- a/src/java/voldemort/store/quota/QuotaUtils.java
+++ b/src/java/voldemort/store/quota/QuotaUtils.java
@@ -16,10 +16,24 @@
 
 package voldemort.store.quota;
 
+import java.io.UnsupportedEncodingException;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
+import voldemort.VoldemortApplicationException;
+import voldemort.server.StoreRepository;
+import voldemort.store.configuration.FileBackedCachingStorageEngine;
+import voldemort.store.system.SystemStoreConstants;
+import voldemort.utils.ByteArray;
+import voldemort.versioning.VectorClock;
+import voldemort.versioning.VectorClockUtils;
+import voldemort.versioning.Versioned;
+
+
 public class QuotaUtils {
+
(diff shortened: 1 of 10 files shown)


src/java/voldemort/store/quota/QuotaUtils.java
```diff
@@ -32,4 +46,47 @@ public static String makeQuotaKey(String storeName, QuotaType quotaType) {
         }
         return quotaTypes;
     }
+
+  private static FileBackedCachingStorageEngine getQuotaStore(StoreRepository repoistory) {
+    FileBackedCachingStorageEngine quotaStore =
+        (FileBackedCachingStorageEngine) repoistory.getStorageEngine(QUOTA_STORE);
```
FelixGV: Typo: ```repoistory```

src/java/voldemort/store/quota/QuotaUtils.java
```diff
@@ -32,4 +46,47 @@ public static String makeQuotaKey(String storeName, QuotaType quotaType) {
...
+
+  private static ByteArray convertToByteArray(String value) {
+    try {
+      return new ByteArray(value.getBytes("UTF8"));
+    } catch (UnsupportedEncodingException ex) {
+      throw new VoldemortApplicationException("Error converting key to byte array " + value, ex);
+    }
+  }
+
+  public static Long getQuota(String storeName, QuotaType type, StoreRepository repoistory) {
```
FelixGV: That's great. It made no sense to use the AdminClient locally on ourselves... (y)

FelixGV:
Cool, thanks! LGTM (:

arunthirupathi:
Addressed the comment and merged it https://github.com/voldemort/voldemort/commit/701e97684ff635d68c79a12f53061250a6cb06c1

PR #402, opened 2016-04-28
Use empty stores for Replace Node CLI test

arunthirupathi:
This should have caught a bug, which is already fixed
in the last commit.

There are still couple of bugs in the Offline Mode.

AdminClient storeOps uses socketPort, which will be down in the
offline mode. AdminPort supports full client operations and hence
it should have used the AdminPort.

AdminClient on the voldemort server uses cluster for bootstrapping
which uses the client port again. This is problematic when the node
is node 0. It should use the Admin Port for bootstrapping.

Those bugs are in backlog will fix them later.

### DIFF
--- a/test/common/voldemort/config/empty-stores.xml
+++ b/test/common/voldemort/config/empty-stores.xml
@@ -1,0 +1,3 @@
+<?xml version="1.0"?>
+<stores>
+</stores>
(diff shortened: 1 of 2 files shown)


FelixGV:
LGTM.

PR #403, opened 2016-05-05
Renamed AdminStoreSwapper#swapStoreData() to #fetchAndSwapStoreData()

FelixGV:
Also clarified one of the logs so that BnP announces that it's about to fetch
(previously, it mentioned "swap" instead).


arunthirupathi:
Looks Good To Me.

FelixGV:
Thanks, I merged it.

PR #404, opened 2016-05-05
Removed 'num.chunks' config parameter from BnP.

FelixGV:
Setting this parameter introduced a subtle failure mode when there were
too few records in the store. It's not worth fixing the other bug since
'num.chunks' is automatically calculated anyway, based on input data size.

This change removes a bit of rope for users to hang themselves with (:

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
@@ -122,7 +122,6 @@ public class VoldemortBuildAndPushJob extends AbstractJob {
     // others.optional
     public final static String KEY_SELECTION = "key.selection";
     public final static String VALUE_SELECTION = "value.selection";
-    public final static String NUM_CHUNKS = "num.chunks";
     public final static String BUILD = "build";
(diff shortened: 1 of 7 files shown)


arunthirupathi:
Looks good to me.

FelixGV:
Thanks, I'll push it in once the tests finish running.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
```diff
@@ -703,7 +702,6 @@ public String runBuildStore(Props props, String url) throws Exception {
                                                             props.getBoolean(SAVE_KEYS, true),
                                                             props.getBoolean(REDUCER_PER_BUCKET, true),
                                                             props.getInt(BUILD_CHUNK_SIZE, 1024 * 1024 * 1024),
-                                                            props.getInt(NUM_CHUNKS, -1),
```
gnb: Instead of never looking up the "num.chunks" key, could you test for it's presence and emit a "The num.chunks parameter is deprecated, ignoring" warning to the Azkaban log?

FelixGV: Done.

PR #405, opened 2016-05-08
Upgrade to BDB JE 5.0.104

arunthirupathi:
This is rebased code for the Pull request
https://github.com/voldemort/voldemort/pull/247

I validated the following scenarios
1) Populate some data with current version ( 5.0.88)
2) Upgrade the BDB je to 5.0.104 and restart the voldemort server
3) Make sure that the data can be read/written.
4) Revert back to 5.0.88 and the operation works as expected.

Unit tests ran and failures were unrelated.

BDB JE 5.0.104 is available as maven artifact and this removes the
last checked in jar from Voldemort. Once this is done, Voldemort
new builds can be published to Maven automatically.


FelixGV:
I am +1 on this.

The previous version of BDB-JE is binary compatible with this one, so if some organization want to be super conservative and stay on the old version, they can still provide Voldemort with the old BDB-JE jar on the classpath and things should work as is. We can call this out in the next release notes.

Meanwhile, this will allow the open-source community to more easily test and deploy Voldemort, and it will bring us one step closer to having a clean build process, with no checked in binaries.

Thanks for doing due diligence on the compatibility test @arunthirupathi !

-F

arunthirupathi:
@FelixGV  thats a good point. At LinkedIn, we are not going to upgrade to 5.0.104 any time soon. We will do exactly what you told and I have verified that code path is unaltered, so I will merge this in.

arunthirupathi:
Now I don't know how the maven automatic publishing works, I believe you have some experience with that, attempt that when you have time.

FelixGV:
Yeah, I can take a stab at the Maven publication one of these days.

PR #406, opened 2016-05-12
Set metadata node

arunthirupathi:
Many minor bug fixes.

1) Make the zoneShrinkageEndToEndTest consistently pass. I noticed that it started failing with a different error message after I changed the ordering of the cluster.xml update and the version update.
2) bootstrap option to get only the store being bootstrapped. Default option is backward compatible, but the option requires newer version of the servers. In LinkedIn we will turn on the option for all use cases.
3) Some other minor bugs I noticed some time back.

### DIFF
--- a/src/java/voldemort/VoldemortClientShell.java
+++ b/src/java/voldemort/VoldemortClientShell.java
@@ -117,7 +117,7 @@ public class VoldemortClientShell {
         try {
             factory = new SocketStoreClientFactory(clientConfig);
             client = factory.getStoreClient(storeName);
-            adminClient = new AdminClient(clientConfig);
+            adminClient = new AdminClient(clientConfig.setIdentifierString("admin"));
 
             storeDef = StoreUtils.getStoreDef(factory.getStoreDefs(), storeName);
 
--- a/src/java/voldemort/client/ClientConfig.java
+++ b/src/java/voldemort/client/ClientConfig.java
@@ -99,6 +99,8 @@ public class ClientConfig {
     protected volatile int maximumTolerableFatalFailures = FailureDetectorConfig.DEFAULT_MAX_TOLERABLE_FATAL_FAILURES;
 
     private volatile int maxBootstrapRetries = 2;
+    private volatile boolean fetchAllStoresXmlInBootstrap = true;
+    private volatile int bootstrapRetryWaitTimeSeconds = 5;
     private volatile String clientContextName = "";
 
     /* 5 second check interval, in ms */
@@ -182,6 +184,8 @@ public class ClientConfig {
     public static final String GETALL_OP_ZONE_AFFINITY = "getall_op_zone_affinity";
     public static final String GETVERSIONS_OP_ZONE_AFFINITY = "getversions_op_zone_affinity";
     public static final String IDENTIFIER_STRING_KEY = "identifier_string";
+    public static final String FETCH_ALL_STORES_XML_IN_BOOTSTRAP = "fetch_all_stores_xml_in_bootstrap";
+    public static final String BOOTSTRAP_RETRY_WAIT_TIME_SECONDS = "bootstrap_retry_wait_time_seconds";
 
     /**
      * Instantiate the client config using a properties file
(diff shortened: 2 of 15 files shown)


arunthirupathi:
@FelixGV  review this when you have time.

src/java/voldemort/client/ClientConfig.java
```diff
@@ -1094,6 +1106,50 @@ public ClientConfig setMaxBootstrapRetries(int maxBootstrapRetries) {
...
+     * 
+     * This property is defaulted to true, so that newer clients can work with
+     * older servers. If there are 100's of stores and you have servers running
+     * on 1.8.1 or above, it is better to set them to false.
+     * 
+     * @param value
+     *            true, to fetch all stores.xml false, to fetch only current
+     *            store
+     */
+    public ClientConfig setFetchAllStoresXmlInBootstrap(boolean value) {
```
FelixGV: Thanks for the great JavaDoc.

Rather than requiring a client config, it'd be great if the client could automatically detect whether the server supports this optimization. Either by trying the optimized path first and falling back on the non-optimized one if the first one fails, or by asking the server for a certain config (we have a generic admin command now that gives visibility into any arbitrary server config, which may be useful).

I don't think this is a blocker. We can leave it like this. I just wanted to mention it because sometimes doing the auto-tuned client does not require that much more code than introducing a manual config.

src/java/voldemort/VoldemortClientShell.java
```diff
@@ -117,7 +117,7 @@ public VoldemortClientShell(ClientConfig clientConfig,
         try {
             factory = new SocketStoreClientFactory(clientConfig);
             client = factory.getStoreClient(storeName);
-            adminClient = new AdminClient(clientConfig);
+            adminClient = new AdminClient(clientConfig.setIdentifierString("admin"));
```
FelixGV: What is the impact of de-registering JMX twice?

I don't want to contest whether we need this. I'm just curious to better understand the original issue.

arunthirupathi: You get an ugly exception when closing the Voldemort client shell but nothing more.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -1156,20 +1210,12 @@ private void internalUpdateMetadataversion(Collection<String> versionKeys) {
...
-                                                    newProperties);
-        }
-
-        /**
-         * Set the metadata versions to the given set
-         * 
-         * @param newProperties The new metadata versions to be set across all
-         *        the nodes in the cluster
-         */
         public void setMetadataversion(Versioned<Properties> newProperties) {
```
FelixGV: Should we refactor this so that the camel casing is consistent across all overloads of this function? (i.e.: ```setMetadataVersion```?)

I think it would also be helpful to move this function next to its other overloads. It's easier to reason about an API when all of the overloads are grouped together...

src/java/voldemort/utils/MetadataVersionStoreUtils.java
```diff
@@ -144,6 +153,7 @@ private static String getPropertiesString(Properties props) {
     public static void setProperties(SystemStoreClient<String, String> versionStore,
```
FelixGV: I haven't pulled down your code so I don't have my IDE to analyze usage, but are we still using this API at all? It looks like might have removed all of its previous usages. If that's the case, I encourage you to get rid of it completely.

src/java/voldemort/VoldemortClientShell.java
FelixGV: Sounds good. Might be worth amending that commit comment to mention this. It's not that important though. As you wish.

FelixGV:
LGTM! Thanks for the commit comments. Since the commits are well broken up, I would suggest not squashing when you merge this in.

src/java/voldemort/utils/MetadataVersionStoreUtils.java
arunthirupathi: I left it on a fall back path, when the node by node set fails. Just to be cautious. We can gut it out after few months may be.

src/java/voldemort/client/protocol/admin/AdminClient.java
arunthirupathi: will fix it.

src/java/voldemort/VoldemortClientShell.java
arunthirupathi: Added a note to the commit

src/java/voldemort/client/ClientConfig.java
arunthirupathi: I agree with you that it can be automatic.
1) Protocol upgrade is a tricky issue  considering Forward/Backward compatibility.
2) Admin Port should ideally be port blocked and opened only from specific boxes, given Voldemort does not have any security built into it.
3) For LinkedIn, I am planning to override the default in the wrapper to do only retrieve one store.

arunthirupathi:
This is merged. Closing it.

PR #408, opened 2016-05-13
Add support for use s3 as input path

xiaozh:


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -220,7 +220,11 @@ public void build() {
             FileSystem tempFs = tempDir.getFileSystem(conf);
             tempFs.delete(tempDir, true);
 
-            long size = sizeOfPath(tempFs, inputPath);
+            //use a list in case we pass a csv path list
```
FelixGV: Is it supported to pass a CSV path list? (I didn't think so, but earlier today you taught me something else which I didn't think Voldemort supported, so who knows eh!)

xiaozh: Please look at : https://aws.amazon.com/blogs/aws/amazon-s3-performance-tips-tricks-seattle-hiring-event/
amazon recommend hash the s3 directory for large inputs. s3 is not really a directory structure but a key/values store.
In this case, we need to create an array contains all the input files instead of the root of path.

PR #410, opened 2016-05-17
Voldemort server cleans up HA state.

FelixGV:
Added code so that the Voldemort server automatically cleans
the shared High Availability state from HDFS when appropriate.

Currently, this new code runs when:
1. An old Read-Only store-version is deleted, which usually
happens asynchronously after a new store-version is activated.
2. When a server transitions from OFFLINE mode to ONLINE mode.

This new benahvior is disabled by default, but can be enabled via:

push.ha.state.auto.cleanup=true

N.B.: This is not tested yet.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
@@ -1,8 +1,11 @@
 package voldemort.store.readonly.swapper;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.Map;
 import java.util.Set;
 
+import com.google.common.collect.Maps;
 import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataOutputStream;
@@ -72,6 +75,7 @@ public class HdfsFailedFetchLock extends FailedFetchLock {
     private final static String RELEASE_LOCK = "release HDFS lock";
     private final static String GET_DISABLED_NODES = "retrieve disabled nodes from HDFS";
     private final static String ADD_DISABLED_NODE = "add a disabled node in HDFS";
+    private final static String CLEAR_OBSOLETE_STATE = "clear obsolete state in HDFS";
     private final static String IO_EXCEPTION = "of an IOException";
     private final static String ALREADY_EXISTS = "it is already acquired (most likely)";
     private final static String UNKNOWN_REASONS = "of unknown reasons";
@@ -330,6 +334,64 @@ public class HdfsFailedFetchLock extends FailedFetchLock {
 
         if (!success) {
             throw new VoldemortException(exceptionMessage(ADD_DISABLED_NODE));
+        }
+    }
+
+    @Override
+    public void removeObsoleteState(int nodeId, String storeName, Map<Long, Boolean> versionToEnabledMap) throws Exception {
+        int attempts = 1;
+        boolean success = false;
+        while (!success && attempts <= maxAttempts) {
+            try {
+                String nodeIdDir = NODE_ID_DIR_PREFIX + nodeId;
+                Path failedStorePath = new Path(clusterDir + "/" + nodeIdDir + "/" + storeName);
+
+                FileStatus[] disabledVersions;
+                try {
+                    disabledVersions = this.fileSystem.listStatus(failedStorePath);
+                    for (FileStatus disabledVersionDir: disabledVersions) {
+                        Long disabledVersion = Long.parseLong(disabledVersionDir.getPath().getName());
+                        Boolean isDisabledLocally = versionToEnabledMap.get(disabledVersion);
+                        if (isDisabledLocally == null || !isDisabledLocally) {
+                            logger.info("The shared state has an obsolete disabled stored version which we will delete: " +
+                                    disabledVersionDir.getPath().toString());
+                            this.fileSystem.delete(disabledVersionDir.getPath(), true);
+                        }
+                    }
+
+                    // Let's see if there's still anything left for this store?
+                    disabledVersions = this.fileSystem.listStatus(failedStorePath);
+                    if (disabledVersions.length == 0) {
+                        logger.info("There are no more disabled versions for this store, so we will delete: " +
+                                failedStorePath.toString());
+                        this.fileSystem.delete(failedStorePath, true);
+                    }
+                } catch (FileNotFoundException e) {
+                    logger.info("The shared state has no obsolete versions in: " + failedStorePath.toString(), e);
+                }
+
+                // Let's see if there's still anything left for this node?
+                Path disabledStoresPath = new Path(clusterDir + "/" + nodeIdDir);
+                try {
+                    FileStatus[] disabledStores = this.fileSystem.listStatus(disabledStoresPath);
+                    if (disabledStores.length == 0) {
+                        logger.info("There are no more disabled stores for this node, so we will delete: " +
+                                disabledStoresPath.toString());
+                        this.fileSystem.delete(disabledStoresPath, true);
+                    }
+                } catch (FileNotFoundException e) {
+                    logger.info("The shared state has no obsolete stores in: " + disabledStoresPath.toString(), e);
+                }
+
+                success = true;
+            }  catch (IOException e) {
+                handleIOException(e, CLEAR_OBSOLETE_STATE, attempts);
+                attempts++;
+            }
+        }
+
+        if (!success) {
+            throw new VoldemortException(exceptionMessage(CLEAR_OBSOLETE_STATE));
         }
     }
 
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
@@ -395,7 +395,7 @@ public class HdfsFailedFetchLock extends FailedFetchLock {
                         this.fileSystem.delete(failedStorePath, true);
                     }
                 } catch (FileNotFoundException e) {
-                    logger.info("The shared state has no obsolete versions in: " + failedStorePath.toString(), e);
+                    logger.info("The shared state has no obsolete versions in: " + failedStorePath.toString());
                 }
 
                 // Let's see if there's still anything left for this node?
@@ -408,7 +408,7 @@ public class HdfsFailedFetchLock extends FailedFetchLock {
                         this.fileSystem.delete(disabledStoresPath, true);
                     }
                 } catch (FileNotFoundException e) {
-                    logger.info("The shared state has no obsolete stores in: " + disabledStoresPath.toString(), e);
+                    logger.info("The shared state has no obsolete stores in: " + disabledStoresPath.toString());
                 }
 
                 success = true;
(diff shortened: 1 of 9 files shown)


src/java/voldemort/store/readonly/StoreVersionManager.java
```diff
@@ -248,6 +256,20 @@ private void removeVersion(long version) {
...
-        // TODO: Sync state with external state (i.e.: FailedFetchLock).
+        removeRemoteObsoleteState();
+    }
+
+    public void removeRemoteObsoleteState() {
+        if (failedFetchLock != null) {
+            try {
+                failedFetchLock.removeObsoleteState(nodeId, storeName, versionToEnabledMap);
+            } catch (Exception e) {
+                logger.error("Failed to execute failedFetchLock.removeObsoleteState()", e);
```
arunthirupathi: Can this logger.error contian the storeName ?

src/java/voldemort/store/readonly/StoreVersionManager.java
```diff
@@ -248,6 +256,20 @@ private void removeVersion(long version) {
...
+            }
+            logger.info("Successfully synced internal state with remote FailedFetchLock state.");
+        } else {
+            logger.info("Will not attempt removeRemoteObsoleteState() because it is disabled. " +
```
arunthirupathi: Can this be in a lower level than info, as it is disabled ?

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -516,6 +521,22 @@ public void goOffline() {
 
     public void goOnline() {
         getMetadataStore().setOfflineState(false);
+        List<StorageEngine<ByteArray, byte[], byte[]>> storageEngines =
+                storageService.getStoreRepository().getStorageEnginesByClass(ReadOnlyStorageEngine.class);
+
+        if (storageEngines.size() > 0) {
```
arunthirupathi: Should this also be done based on the voldemortConfig to clean up the state ?

src/java/voldemort/store/readonly/ReadOnlyStorageConfiguration.java
```diff
@@ -48,7 +49,18 @@ public ReadOnlyStorageConfiguration(VoldemortConfig config) {
         this.nodeId = config.getNodeId();
         this.deleteBackupMs = config.getReadOnlyDeleteBackupMs();
         this.maxValueBufferAllocationSize = config.getReadOnlyMaxValueBufferAllocationSize();
-        this.voldConfig = config;
+
+        FailedFetchLock failedFetchLockInstance;
+        if (config.getHighAvailabilityStateAutoCleanUp()) {
+            try {
+                failedFetchLockInstance = FailedFetchLock.getLock(config, null);
```
arunthirupathi: HdfsFailedFetchLock initializes the HadoopFileSystem on the constructor. The HadoopFileSystem opened on the constructor is prone to connection resets and failures. Also this will force all the store to have an open handle to the HdfsFileSystem which could cause more issues.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
```diff
@@ -333,6 +337,64 @@ public void addDisabledNode(int nodeId,
...
+                        this.fileSystem.delete(disabledStoresPath, true);
+                    }
+                } catch (FileNotFoundException e) {
+                    logger.info("The shared state has no obsolete stores in: " + disabledStoresPath.toString(), e);
+                }
+
+                success = true;
+            }  catch (IOException e) {
+                handleIOException(e, CLEAR_OBSOLETE_STATE, attempts);
+                attempts++;
```
gnb: No delay before retry?

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -148,6 +148,7 @@
     public static final String PUSH_HA_LOCK_PATH = "push.ha.lock.path";
     public static final String PUSH_HA_LOCK_IMPLEMENTATION = "push.ha.lock.implementation";
     public static final String PUSH_HA_MAX_NODE_FAILURES = "push.ha.max.node.failure";
+    public static final String PUSH_HA_STATE_AUTO_CLEANUP = "push.ha.state.auto.cleanup";
```
gnb: I think it would be better to prove that deleting the state is the right thing to do, than to put the decision behind a configuration variable.  That's just delegating the decision to an SRE who has less time and context to work out the right thing to do, and already has a manual workaround.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -516,6 +521,22 @@ public void goOffline() {
...
+            for (StorageEngine storageEngine: storageEngines) {
+                StoreVersionManager storeVersionManager =
+                        (StoreVersionManager) storageEngine.getCapability(StoreCapabilityType.DISABLE_STORE_VERSION);
+                try {
+                    storeVersionManager.removeRemoteObsoleteState();
+                } catch (Exception e) {
+                    logger.error("Failed to removeRemoteObsoleteState() for store: " + storageEngine.getName(), e);
+                }
+            }
+        }
```
gnb: So much for information hiding.  Why not add a `public void goingOnline(void) throws Exception` method to class `StorageEngine`, whose default implementation does nothing?

arunthirupathi: Also on looking at this again, it does store by store. For LSG1 with 70 stores, it takes about 60 seconds to run on each node. With 30 nodes, it takes 30 minutes to set the cluster to online.

Can we do a bulk operation of deleting a node-id ?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/swapper/HdfsFailedFetchLock.java
FelixGV: ```handleIOException``` does wait.

src/java/voldemort/server/VoldemortServer.java
FelixGV: I think that there is an opportunity for optimization, but I'm not sure that it's worth it, since this is a rare operation.

I don't understand how you arrive at your 30 minutes figure. We only need to bring back one node from offline to online, not 30.

src/java/voldemort/store/readonly/ReadOnlyStorageConfiguration.java
FelixGV: Sure, I could create a new ```FailedFetchLock``` and then close it every time we need one, rather than hanging on to it.

src/java/voldemort/server/VoldemortServer.java
FelixGV: The check already happens deeper in the call stack, but I can put the check in both places if that's deemed cleaner.

src/java/voldemort/store/readonly/StoreVersionManager.java
FelixGV: Yes.

src/java/voldemort/store/readonly/StoreVersionManager.java
FelixGV: Yes.

src/java/voldemort/server/VoldemortServer.java
arunthirupathi: If you set all nodes to online, it will take 30 minutes. You are right, we are doing only one node at a time.

But even then it feels too much time to wait, when a human is running the operation.

src/java/voldemort/server/VoldemortConfig.java
FelixGV: Well... do you have any suggestions for how we can prove that it is the right thing to do besides actually trying it? Once battle-tested and proven, I'm totally fine with changing this setting's default to true.

src/java/voldemort/server/VoldemortServer.java
FelixGV: I changed this to be asynchronous in my latest commit.

src/java/voldemort/store/readonly/ReadOnlyStorageConfiguration.java
FelixGV: My latest commit closes the ```FailedFetchLock``` and its underlying Hadoop FS as soon as it's finished with it. It also instantiates just one ```FailedFetchLock``` to deal with all stores in the offline to online transition, so the memory footprint should be minimal.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -520,21 +524,60 @@ public void goOffline() {
...
-                    logger.error("Failed to removeRemoteObsoleteState() for store: " + storageEngine.getName(), e);
-                }
-            }
+        if (getVoldemortConfig().getHighAvailabilityStateAutoCleanUp()) {
+            final int asyncRequestId = asyncService.getUniqueRequestId();
+            logger.info("Since " + VoldemortConfig.PUSH_HA_STATE_AUTO_CLEANUP + "=true, async task ID " +
+                        asyncRequestId + " will be launched to ensure the HA state is cleaned up.");
+            asyncService.submitOperation(
+                    asyncRequestId,
+                    new AsyncOperation(asyncRequestId, "HA state clean up.") {
```
arunthirupathi: Can you please create a new class for this , instead of this anonymous class ?

arunthirupathi:
Left some minor comments, looks good to me.

src/java/voldemort/server/VoldemortServer.java
FelixGV: Sure, I can. It still seems contentious whether this should be asynchronous in the first place or not, however. Let's settle this before making any further changes.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -520,24 +524,84 @@ public void goOffline() {
...
+                    if (firstItem) {
+                        firstItem = false;
+                    } else {
+                        stringBuilder.append(", ");
+                    }
+                    stringBuilder.append(storeName);
+
+                }
+                logger.warn(stringBuilder.toString());
+                return ReadOnlyStoreStatusValidation.SOME_STORES_DISABLED;
```
arunthirupathi: I would prefer to throw an exception with message string you built here. It is hard for a SRE to hop on the box and grep for log message to find the error.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -414,7 +414,7 @@ public void run() {
                     Utils.rm(file);
                     logger.info("Deleting of " + file.getAbsolutePath()
                                 + " completed successfully.");
-                    storeVersionManager.syncInternalStateFromFileSystem();
+                    storeVersionManager.syncInternalStateFromFileSystem(true);
```
arunthirupathi: In my opinion, it would have been easier to add an another method call here, to remove the HDFS state, instead of the boolean flag.

This is a nitpick, so no need to change the code.

src/java/voldemort/server/VoldemortServer.java
FelixGV: Yeah, makes sense. I changed it once more to do that.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -524,26 +524,28 @@ public void goOffline() {
...
-            case NO_STORES_DISABLED:
-                createOnlineServices();
-                startOnlineServices();
-                break;
+        ReadOnlyStoreStatusValidation validation = validateReadOnlyStoreStatusBeforeGoingOnline();
+
+        if (validation.readyToGoOnline) {
+            createOnlineServices();
+            startOnlineServices();
+            getMetadataStore().setOfflineState(false);
```
arunthirupathi: In the existing code, this was the first statement. Any reason, why you are changing the order ?

If not, I would like to keep the getMetadataStore().setOfflineState(false) to be the first statement.

arunthirupathi:
When you are done testing this, please merge this in.

arunthirupathi:
Closed by mistake, re-opened it.

FelixGV:
I'm done testing. I squashed and merged.

PR #411, opened 2016-05-20
Last commit regressed the Hadoop Fetcher

arunthirupathi:
There were tests failing, and on investigation it turned out that
negation was missed.

My preference was to write == false instead of negation, but that is C++
style.


squarY:
LGTM. Sorry for that.

PR #412, opened 2016-05-23
Lower the node unavailable error to debug

arunthirupathi:
When async connect fails, selector reports the error and it is cached in
memory. Next connect call will get the error if it happens within twice
the timeout period.

The logs are logged at the info level, which spams the logs when the
server is unavailable for extended period of times. Now it is dropped
down to debug level.

### DIFF
--- a/src/java/voldemort/utils/pool/KeyedResourcePool.java
+++ b/src/java/voldemort/utils/pool/KeyedResourcePool.java
@@ -557,9 +557,10 @@ public class KeyedResourcePool<K, V> {
                 skippedExceptionCount++;
                 if(elapsedTime <= excpetionReportTimeMS) {
                     Exception e = entry.getSecond();
-                    logger.info(" Throwing remembered exception. time elapsed (ms) " + elapsedTime
-                                + ". Exception : "
-                                + e.getMessage());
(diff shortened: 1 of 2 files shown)


src/java/voldemort/utils/pool/KeyedResourcePool.java
```diff
@@ -557,9 +557,10 @@ private void throwReportedExceptions() throws Exception {
                 skippedExceptionCount++;
                 if(elapsedTime <= excpetionReportTimeMS) {
                     Exception e = entry.getSecond();
-                    logger.info(" Throwing remembered exception. time elapsed (ms) " + elapsedTime
-                                + ". Exception : "
-                                + e.getMessage());
+                    if(logger.isDebugEnabled()) {
```
gaojieliu: From the name of this class, it could be used by multiple places, so is it possible that we still need info logging in other scenarios?

arunthirupathi: Though it could be used at many places, there is one mainline use and others are there, deprecated or for testing purposes.

gaojieliu:
Looks good!

PR #413, opened 2016-06-01
Idle connection timeout

arunthirupathi:
1) Add idle connection timeout support to Voldemort.
2) Reduce the socket timeout ( timeout once socket.read() is called from 24 hours to 1 minute)
3) Minor fixes to remove Properties(properties)
4) Debug information fix.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
@@ -49,6 +49,7 @@ import org.apache.log4j.Logger;
 import voldemort.VoldemortException;
 import voldemort.client.ClientConfig;
 import voldemort.client.protocol.admin.AdminClient;
+import voldemort.client.protocol.admin.AdminClientConfig;
 import voldemort.client.protocol.pb.VAdminProto;
 import voldemort.cluster.Cluster;
 import voldemort.cluster.Node;
@@ -194,9 +195,11 @@ public class VoldemortBuildAndPushJob extends AbstractJob {
                 if (clusterURLs.contains(url))
                     throw new VoldemortException("the URL: " + url + " is duplicated. Please check it out.");
                 this.clusterURLs.add(url);
-                AdminClient adminClient = new AdminClient(new ClientConfig().setBootstrapUrls(url)
-                                                                            .setConnectionTimeout(15,
-                                                                                                  TimeUnit.SECONDS));
+                ClientConfig config = new ClientConfig().setBootstrapUrls(url)
+                                    .setConnectionTimeout(15,TimeUnit.SECONDS);
+                
+                AdminClientConfig adminConfig = new AdminClientConfig().setAdminSocketTimeoutSec(60);
+                AdminClient adminClient = new AdminClient(adminConfig, config);
                 this.adminClientPerCluster.put(url, adminClient);
                 this.closeables.add(adminClient);
             }
--- a/src/java/voldemort/client/ClientConfig.java
+++ b/src/java/voldemort/client/ClientConfig.java
@@ -51,6 +51,7 @@ public class ClientConfig {
     private volatile int maxConnectionsPerNode = 50;
     private volatile long connectionTimeoutMs = 500;
     private volatile long socketTimeoutMs = 5000;
+    private volatile long idleConnectionTimeoutMs = -1;
     private volatile boolean socketKeepAlive = false;
     private volatile int selectors = 8;
     private volatile long routingTimeoutMs = 5000;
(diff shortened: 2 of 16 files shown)


src/java/voldemort/client/ClientConfig.java
```diff
@@ -571,6 +577,52 @@ public ClientConfig setSocketTimeout(int socketTimeout, TimeUnit unit) {
...
+        return unit.convert(idleConnectionTimeoutMs, TimeUnit.MILLISECONDS);
+    }
+
+    /**
+     * Set the timeout for idle connections. Voldemort client caches all
+     * connections to the Voldemort server. This setting allows the a connection
+     * to be dropped, if it is idle for more than this time.
+     * 
+     * This could be useful in the following cases 1) Voldemort client is not
+     * directly connected to the server and is connected via proxy of firewall.
```
stotch: Just to avoid confusion ...
"proxy of firewall"
should be ...
"a proxy or firewall"

src/java/voldemort/client/ClientConfig.java
```diff
@@ -571,6 +577,52 @@ public ClientConfig setSocketTimeout(int socketTimeout, TimeUnit unit) {
...
+     * @return ClientConfig object for chained set
+     * 
+     *         throws {@link IllegalArgumentException} if the timeout is greater
+     *         than 0, but less than 10 minutes.
+     */
+    public ClientConfig setIdleConnectionTimeout(long idleConnectionTimeout, TimeUnit unit) {
+        if (idleConnectionTimeout <= 0) {
+            this.idleConnectionTimeoutMs = -1;
+        } else {
+            if(unit.toMinutes(idleConnectionTimeout) < 10) {
```
stotch: Why are we limiting this to 10 min?

src/java/voldemort/utils/MetadataVersionStoreUtils.java
```diff
@@ -80,7 +80,12 @@ public static Properties mergeVersions(Properties prop1, Properties prop2) {
             return prop1;
         }
 
-        Properties result = new Properties(prop1);
```
FelixGV: Why not use Voldemort's `Props` class instead of Java `Properties`? Voldemort's implementation already supports this merging use case. Anyway, no need to do any major code overhaul as part of this PR, but it may be worth at least marking this `mergeVersions` function as `@Deprecated` since we should probably avoid using Java `Properties`.

src/java/voldemort/client/ClientConfig.java
```diff
@@ -571,6 +577,52 @@ public ClientConfig setSocketTimeout(int socketTimeout, TimeUnit unit) {
...
+     *            zero or negative number to disable the feature ( default -1)
+     *            timeout
+     * @param unit
+     *            {@link TimeUnit}
+     * @return ClientConfig object for chained set
+     * 
+     *         throws {@link IllegalArgumentException} if the timeout is greater
+     *         than 0, but less than 10 minutes.
+     */
+    public ClientConfig setIdleConnectionTimeout(long idleConnectionTimeout, TimeUnit unit) {
```
FelixGV: Thanks for the detailed JavaDoc. Can we add a line mentioning that this config is only (potentially) relevant to very low throughput clients? Also, a small comment on why the 10 minutes minimum would be great.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
```diff
@@ -719,7 +722,7 @@ public String runBuildStore(Props props, String url) throws Exception {
 
     public void runPushStore(Props props, String url, String dataDir) throws Exception {
         // For backwards compatibility http timeout = admin timeout
-        int httpTimeoutMs = 1000 * props.getInt(PUSH_HTTP_TIMEOUT_SECONDS, 24 * 60 * 60);
+        int httpTimeoutMs = 1000 * props.getInt(PUSH_HTTP_TIMEOUT_SECONDS, 60);
```
FelixGV: This will break the `AdminStoreSwapper`, because it also uses the same parameter for the total fetch time. For total fetch time, 24 hours is not necessarily unreasonable (unlimited may be fine too). The problem is that we use the same config parameter for both of these settings. We should either stop using this parameter in the `AdminStoreSwapper` (and then I guess we may want to default to -1 for unlimited fetch time) or we could create a new setting so that these two are controlled separately. I'll leave it up to you which of these two options you prefer, but this cannot be merged in without either one of them.

FelixGV:
LGTM, except for the comment I made in BnP. That one's a blocker, I think.

src/java/voldemort/utils/MetadataVersionStoreUtils.java
arunthirupathi: I believe Props we are using mostly from the Config perspective. We are using the Properties for  our versions serializing to byte array and loading back from byte array. marking mergeVersions as deprecated is not an option without the alternative.

src/java/voldemort/client/ClientConfig.java
arunthirupathi: will do.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
arunthirupathi: Fixed it, thanks for the great catch. I should have been careful while modifying it.

PR #415, opened 2016-06-09
Quota type

arunthirupathi:
1) NPE fix in the BnP
2) API consistency for QuotaType.
3) Quota set, tries all nodes before rethrowing the last exception. ( Consistent with verifyOrAddStore behavior, where it skips the failed nodes while adding the store).

### DIFF
--- a/src/java/voldemort/store/StoreDefinition.java
+++ b/src/java/voldemort/store/StoreDefinition.java
@@ -510,15 +510,15 @@ public class StoreDefinition implements Serializable {
                 addToDiff("Preferred writes", this.getPreferredWrites(), other.getPreferredWrites(), sb, thisName, otherName);
             }
             if (!this.getKeySerializer().equals(other.getKeySerializer())) {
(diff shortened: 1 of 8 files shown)


src/java/voldemort/store/StoreDefinition.java
```diff
@@ -573,6 +573,10 @@ private void addToDiff(String propertyName, long thisValue, long otherValue, Str
     private void addToDiff(String propertyName, int thisValue, int otherValue, StringBuilder sb, String thisName, String otherName) {
         addToDiff(propertyName, Integer.toString(thisValue), Integer.toString(otherValue), sb, thisName, otherName);
     }
+    
+    private void addToDiff(String propertyName, Object thisValue, Object otherValue, StringBuilder sb, String thisName, String otherName) {
```
FelixGV: Would you mind using `String.valueOf` at the call-site rather than changing the signature? It seems cleaner to have a strong type in the function signature. Not a big deal either way. Just a slight preference.

FelixGV:
Thanks for making these changes! LGTM.

PR #416, opened 2016-06-11
Logging changes

arunthirupathi:
Logging changes when store is missing and BnP directory processing.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
@@ -22,8 +22,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import com.google.common.collect.Lists;
+
 import org.apache.avro.Schema;
 import org.apache.avro.mapred.AvroJob;
 import org.apache.avro.mapred.AvroOutputFormat;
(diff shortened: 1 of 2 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -511,8 +539,6 @@ public boolean accept(Path arg0) {
             }
 
             long diskSizeForNodeInBytes = dataSizeInBytes + indexSizeInBytes;
-            logger.info(directoryName + ": Checksum = " + checkSum +
```
FelixGV: Can we keep this as a debug statement?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -348,11 +372,9 @@ public void build() {
                 Path directoryPath = new Path(outputDir.toString(), directoryName);
 
                 if(!outputFs.exists(directoryPath)) {
-                    logger.info("No data generated for " + directoryName
```
FelixGV: Can we keep this as a debug statement?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -348,11 +372,9 @@ public void build() {
...
-                                        + ". Generating empty folder");
+                    emptyDirectories.add(directorySuffix);
                     outputFs.mkdirs(directoryPath); // Create empty folder
                     outputFs.setPermission(directoryPath, new FsPermission(HADOOP_FILE_PERMISSION));
-                    logger.info("Setting permission to 755 for " + directoryPath);
```
FelixGV: Can we keep this as a debug statement? Perhaps merged into the other statement above, if you'd like.

src/java/voldemort/store/metadata/MetadataStore.java
```diff
@@ -531,31 +531,34 @@ public Object getCapability(StoreCapabilityType capability) {
...
+                } catch(Exception e) {
+                    throw new VoldemortException("Failed to read metadata key:"
+                                                         + ByteUtils.getString(keyBytes.get(),
+                                                                               "UTF-8")
+                                                         + " delete config/.temp config/.version directories and restart.",
+                                                 e);
+                }
             } else {
-                throw new VoldemortException("Unhandled Key:" + key + " for MetadataStore get()");
+                throw new VoldemortException("Store " + key + " does not exist on node "
```
FelixGV: Is this code only used for "stores" or are there other types of metadata that use the same code? If it's the latter, the more specific message may not make sense in all contexts, and it may be preferable to override the message higher up the call stack.

I don't know, I haven't looked at that code in detail.

FelixGV:
I left some minor comments. LGTM overall.

src/java/voldemort/store/metadata/MetadataStore.java
arunthirupathi: You are right, the code could be exercised in 2 paths.

1) client Bootstrap, when it tries to do the getStoreClient call.
2) during vadmin.sh meta get <key> 

the second one will throw an confusing exception, but only SRE should run it and first one is more important for our partners to understand.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
arunthirupathi: will do.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
arunthirupathi: will do.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
arunthirupathi: will do.

src/java/voldemort/store/metadata/MetadataStore.java
FelixGV: Ok but... can't we handle the situation higher up the call-stack, where we would know that the request received is a client bootstrap? Or is the client bootstrap and the `vadmin.sh meta get` indistinguishable from the server's perspective?

arunthirupathi: They are indistinguishable from the server. They are both part of getRemoteMetadata.

PR #422, opened 2016-06-23
Increase the timeout of vadmin tool

arunthirupathi:
Increased the timeout of Vadmin tool to 5 seconds from 500 ms default.


FelixGV:
LGTM.

PR #423, opened 2016-06-24
Admin client fixes

arunthirupathi:
1) verifyOrAddStore option to retireve only single store, instead of fetching all stores, which on some clusters with large number of stores , take a long amount of time.
2) a way to check is it easy to re-use the AdminClient for a longer period of time.

### DIFF
--- a/src/java/voldemort/client/protocol/admin/AdminClient.java
+++ b/src/java/voldemort/client/protocol/admin/AdminClient.java
@@ -86,6 +86,7 @@ import voldemort.store.ErrorCodeMapper;
 import voldemort.store.InvalidMetadataException;
 import voldemort.store.Store;
 import voldemort.store.StoreDefinition;
+import voldemort.store.StoreNotFoundException;
 import voldemort.store.StoreUtils;
 import voldemort.store.UnreachableStoreException;
 import voldemort.store.metadata.MetadataStore;
@@ -169,6 +170,7 @@ public class AdminClient implements Closeable {
     private final NetworkClassLoader networkClassLoader;
     private final AdminClientConfig adminClientConfig;
     private final long clusterVersion;
+    private final boolean fetchSingleStore;
     private final String debugInfo;
     private Cluster currentCluster;
     private SystemStoreClient<String, String> metadataVersionSysStoreClient = null;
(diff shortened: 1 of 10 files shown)


src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -299,6 +301,28 @@ public void close() {
...
     /**
+     * If Cluster is modified ( nodes are added/removed), using the old
+     * AdminClient will cause inconsistent operations. This method returns true
+     * under such cases, so that caller can discard the current AdminClient and
+     * create new AdminClient.
+     * 
+     * @return true if cluster is modified since the AdminClient is create,
+     *         false otherwise
+     */
+    public boolean isClusterModified() {
```
FelixGV: I don't think that we should expose leak this complexity to the code which uses the `AdminClient`.

The `ZenStoreClient` refreshes its `cluster.xml` on its own. Why can't the `AdminClient` do the same?

FelixGV:
I think the `verifyOrAddStore` change is a nice optimization. Feel free to merge that one in right away.

The other commit, I'm not so sure... Can you elaborate on why this is needed and why can't we leverage the same thing that we're already doing for the StoreClients?

FelixGV:
Also, we should modify BnP to use the optimization, since BnP does not need to query all stores.

src/java/voldemort/client/protocol/admin/AdminClient.java
arunthirupathi: For AdminCleint to be self sufficient, it needs to do the following
1) Refresh cluster.xml
2) Flush all the connections and reset all its state to new one.
3) Some of the ongoing operation, may need to be re-attempted or retried.

ZenStoreClient does that to cope with cluster replacement and cluster failure. But AdminClient is designed with the assumption that a cluster must be healthy or the command will fail.

Doing the above 3 is very tough if not impossible. Exposing the complexity to the caller is compromise, but will work in the short term. AdminClient is designed for operator and making it run on a self recovering API environment is a very big task.

Thoughts/Opinions ?

arunthirupathi:
About the verifyOrAddStore fix. ZenStoreClient retrieves the metadata using different protocol/port and AdminClient uses a different AdminCommand (GET_METADATA). Can both of them do the same thing , probably not . Admin Port is not accessible for clients generally and client ports can be down on an offline server state. So it is little hard to leverage one for the other and probably they should stay independent.

arunthirupathi:
Amended my last commit to add one more test, to make sure that AdminClient works across node restart.

PR #426, opened 2016-07-11
Admin client fixes

arunthirupathi:
Fixed a log message when fetch is disabled.

Introduced a new getAll api for retrieving multiple quotas at the same time. Tests to cover the same have been added as well.

### DIFF
--- a/src/java/voldemort/client/protocol/admin/AdminClient.java
+++ b/src/java/voldemort/client/protocol/admin/AdminClient.java
@@ -3333,8 +3333,6 @@ public class AdminClient implements Closeable {
             }
             return result;
         }
-
-        // As needed, add 'getall', and so on interfaces...
     }
 
     /**
--- a/src/java/voldemort/client/protocol/admin/AdminClient.java
+++ b/src/java/voldemort/client/protocol/admin/AdminClient.java
@@ -176,7 +176,6 @@ public class AdminClient implements Closeable {
     private final String debugInfo;
     private Cluster currentCluster;
     private SystemStoreClient<String, String> metadataVersionSysStoreClient = null;
-    private SystemStoreClient<String, String> quotaSysStoreClient = null;
     private SystemStoreClientFactory<String, String> systemStoreFactory = null;
 
     final public AdminClient.HelperOperations helperOps;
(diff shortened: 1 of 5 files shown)


src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1054,14 +1054,12 @@ private String swapStore(String storeName, String directory) throws VoldemortExc
                                                                                                             .setStatus("started");
         try {
             if(!metadataStore.getReadOnlyFetchEnabledUnlocked()) {
-                throw new ReadOnlyFetchDisabledException("Read-only fetcher is disabled in "
-                                             + metadataStore.getServerStateUnlocked()
-                                             + " state on node " + metadataStore.getNodeId());
+                throw new ReadOnlyFetchDisabledException("Pushes to this node have been administratively disabled."
```
FelixGV: I don't think that this message is entirely better... `metadataStore.getReadOnlyFetchEnabledUnlocked()` will return false when a node is in `OFFLINE` state, right? This could be triggered automatically by BnP HA, so it does not reflect an "administrative" intervention. I think it would be confusing to state that in the error message.

Furthermore, I think it's useful to have the server state and node ID as part of the log message. You can still add the next line ("Please reach out to the cluster admin for assistance") but I think the beginning of the message should not be changed.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -3305,6 +3300,22 @@ public void putNodeKeyValue(String storeName, NodeValue<ByteArray, byte[]> nodeK
...
+         * Fetch values for given keys on a specific store present on a specific node.
+         * 
+         * @param storeName Name of the Store
+         * @param nodeId Id of the node to query from
+         * @param keys List of keys to be required.
+         * @return
+         * Only keys that has values will be returned in a Map.
+         * Keys that does not exists will not be present in the Map.
+         */
+        public Map<ByteArray, List<Versioned<byte[]>>> getAllNodeKeys(String storeName, int nodeId,
```
FelixGV: `getAllNodeKeys` sounds a bit confusing to me... Can we call it `getAllKeys` or `getAll`? Or if we dare move away from Voldemort's traditionally confusing names, perhaps even `multiGet`?

FelixGV:
The multi-get for system stores LGTM. I think we should reconsider the error message change when fetches are disabled.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
arunthirupathi: You are right, BnP HA can trigger this as well. I think  hostname is already prefix on the BnP logging so id is a duplicate information.

When BnP HA is activated, the job usually succeeds, so we hardly see this error logs. This is important, when we are doing rolling upgrades when the job usually fails. For that use case, the new error message is better suited.

FelixGV: Sure. I think it's fine to leave the second line of your new error message. But I definitely think we should not make the error message over-specific by claiming that this is due to an administrative operation when it is not necessarily the case.

src/java/voldemort/client/protocol/admin/AdminClient.java
arunthirupathi: this is voldemort convention. we have getNodeKey, putNodeKeyValue . deleteNodeKeyValue (although it should be deleteNodeKey). I followed the convention and named it getAllNodeKeys. 

I agree with you that mulitGet is better name. But if we break away from the convention it would look more confusing than it solves.

FelixGV: Ok, I wasn't aware of the convention with the other APIs. We can leave it as is then.

arunthirupathi:
The commits are merged in.

PR #427, opened 2016-07-11
Auto detect node

arunthirupathi:
Voldemort server takes node id as configuration parameters.
It relies on the node id to identify its role in the cluster.
But most production deployments has only one voldemort server
per host in a cluster. Under these conditions the deduction
of node ids can be automated.

All Configuration is disabled by default, but validation is enforced for the cluster update and the node id update.

### DIFF
--- a/config/single_node_cluster/config/server.properties
+++ b/config/single_node_cluster/config/server.properties
@@ -1,6 +1,5 @@
-# The ID of *this* particular cluster node
-node.id=0
-
+# Enable auto detection of node id
+enable.node.id.detection=true
 max.threads=100
 
 ############### DB options ######################
--- a/src/java/voldemort/common/nio/SelectorManagerWorker.java
+++ b/src/java/voldemort/common/nio/SelectorManagerWorker.java
@@ -106,11 +106,9 @@ public abstract class SelectorManagerWorker implements Runnable {
             else if(selectionKey.isWritable())
                 write(selectionKey);
             else if(!selectionKey.isValid())
-                throw new IllegalStateException("Selection key not valid for "
-                                                + socketChannel.socket());
+                throw new IllegalStateException("Selection key not valid for " + getDebugInfo());
             else
-                throw new IllegalStateException("Unknown state, not readable, writable, or valid for "
-                                                + socketChannel.socket());
+                throw new IllegalStateException("Unknown state, not readable, writable, or valid for " + getDebugInfo());
         } catch(ClosedByInterruptException e) {
             reportException(e);
             close();
@@ -122,12 +120,11 @@ public abstract class SelectorManagerWorker implements Runnable {
             reportException(e);
             close();
         } catch(IOException e) {
-            logger.info("Connection reset from " + socketChannel.socket() + " with message - "
-                        + e.getMessage());
+            logger.info("Connection reset from " + getDebugInfo() + " with message - " + e.getMessage());
             reportException(e);
             close();
         } catch(Throwable t) {
-            logger.error("Caught throwable from " + socketChannel.socket(), t);
+            logger.error("Caught throwable from " + getDebugInfo(), t);
             close();
         }
     }
@@ -208,6 +205,10 @@ public abstract class SelectorManagerWorker implements Runnable {
                      + inputStream.getBuffer().limit() + ", remaining: "
                      + inputStream.getBuffer().remaining() + ", capacity: "
                      + inputStream.getBuffer().capacity() + " - for " + socketChannel.socket());
+    }
+
+    protected String getDebugInfo() {
+        return String.valueOf(socketChannel.socket());
     }
 
 }
--- a/src/java/voldemort/server/Host.java
+++ b/src/java/voldemort/server/Host.java
@@ -1,0 +1,26 @@
+package voldemort.server;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import voldemort.VoldemortApplicationException;
+
+
+public class Host {
+
+    private static InetAddress getLocal() {
+        try {
+            return InetAddress.getLocalHost();
+        } catch(UnknownHostException ex) {
+            throw new VoldemortApplicationException("Error retrieving local host", ex);
+        }
+    }
+
+    public static String getName() {
+        return getLocal().getHostName();
+    }
+
+    public static String getFQDN() {
+        return getLocal().getCanonicalHostName();
+    }
+}
--- a/src/java/voldemort/server/HostMatcher.java
+++ b/src/java/voldemort/server/HostMatcher.java
@@ -1,0 +1,66 @@
+package voldemort.server;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+
+import voldemort.VoldemortApplicationException;
+import voldemort.cluster.Node;
+import voldemort.utils.ReflectUtils;
+
+
+public abstract class HostMatcher {
+
+    private final List<String> types;
+    public static final String HOSTNAME = "hostname";
+    public static final String FQDN = "fqdn";
+
+    public HostMatcher(List<String> types) {
+        this.types = types;
+    }
+
+    public boolean match(Node node) {
+        for(String type: types) {
+            String host = getHost(type);
+            if(match(node, host)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    abstract protected boolean match(Node node, String host);
+
+    protected String getHost(String type) {
+        if(type.equalsIgnoreCase(HOSTNAME)) {
+            return Host.getName();
+        } else if(type.equalsIgnoreCase(FQDN)) {
+            return Host.getFQDN();
+        } else {
+            throw new VoldemortApplicationException("Unrecognized type" + type);
+        }
+    }
+
+    public static HostMatcher getImplementation(String className, List<String> types) {
+        if(className == PosixHostMatcher.class.getName()) {
+            return new PosixHostMatcher(types);
+        } else {
+            Class<?> factoryClass = ReflectUtils.loadClass(className.trim());
+            return (HostMatcher) ReflectUtils.callConstructor(factoryClass,
+                                                              new Class<?>[] { types.getClass() },
+                                                              new Object[] { types });
+        }
+    }
+
+    public boolean isLocalAddress(Node node) {
+        String host = node.getHost();
+        InetAddress hostAddress;
+        try {
+            hostAddress = InetAddress.getByName(host);
+        } catch(UnknownHostException ex) {
+            throw new VoldemortApplicationException("Error retrieving InetAddress for host " + host,
+                                                    ex);
+        }
+        return hostAddress.isAnyLocalAddress() || hostAddress.isLoopbackAddress();
+    }
+}
--- a/src/java/voldemort/server/HostMatcher.java
+++ b/src/java/voldemort/server/HostMatcher.java
@@ -2,6 +2,7 @@ package voldemort.server;
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.Arrays;
 import java.util.List;
 
 import voldemort.VoldemortApplicationException;
@@ -63,4 +64,15 @@ public abstract class HostMatcher {
         }
         return hostAddress.isAnyLocalAddress() || hostAddress.isLoopbackAddress();
     }
+
+    public String getDebugInfo() {
+        String debugInfo = "HostMatcher [types=" + Arrays.toString(types.toArray()) + "].";
+        for(String type: types) {
+            String host = getHost(type);
+            debugInfo += " Type : " + type;
+            debugInfo += " , Value " + host + " .";
+        }
+        return debugInfo;
+    }
+
 }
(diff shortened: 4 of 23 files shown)


config/single_node_cluster/config/server.properties
```diff
@@ -1,6 +1,5 @@
-# The ID of *this* particular cluster node
-node.id=0
-
+# Enable auto detection of node id
+enable.node.id.detection=true
```
gnb: I like this config variable name.

src/java/voldemort/server/Host.java
```diff
@@ -0,0 +1,26 @@
...
+        }
+    }
+
+    public static String getName() {
+        return getLocal().getHostName();
+    }
+
+    public static String getFQDN() {
+        return getLocal().getCanonicalHostName();
+    }
```
gnb: A reading of the JDK code shows that for an `InetAddress` object of the shape returned by `InetAddress.getLocalHost()`, both `getHostName()` and `getCanonicalHostName()` always return exactly the same thing, except that the latter takes 6 trips through the platform naming system instead of 4.  So you could replace this entire class with a call to `InetAddress.getLocalHost().getHostName()`.

src/java/voldemort/server/HostMatcher.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+
+import voldemort.VoldemortApplicationException;
+import voldemort.cluster.Node;
+import voldemort.utils.ReflectUtils;
+
+
+public abstract class HostMatcher {
+
+    private final List<String> types;
```
gnb: Given comments above, this `types` member and all the code that uses it is completely unnecessary.

src/java/voldemort/server/PosixHostMatcher.java
```diff
@@ -0,0 +1,19 @@
+package voldemort.server;
+
+import java.util.List;
+
+import voldemort.cluster.Node;
+
+
+public class PosixHostMatcher extends HostMatcher {
```
gnb: Why does this class exist?  Under what conceivable circumstances would the hostname comparison mechanism be anything except the simple and obvious `node.getHost().equals(host)` you have here?  FQDNs are an *Internet* concept not a POSIX concept, they're not going to be different on Windows.  The `match()` method should be moved into the parent class which should be made concrete.

src/java/voldemort/server/NodeIdUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        if(matches.isEmpty()) {
+            throw new VoldemortApplicationException(" No nodes in the cluster matched the current node "
+                                                + Arrays.toString(cluster.getNodes().toArray()));
+        } else if ( matches.size() > 1) {
+            throw new VoldemortApplicationException(" More than one node matched "
+                                                    + Arrays.toString(matches.toArray()));
+        } else {
+            logger.info(" computed node Id match successfully " + matches.get(0).briefToString());
```
gnb: I look forward to seeing this in logs.

src/java/voldemort/server/NodeIdUtils.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        if(!errors.isEmpty()) {
+            if(errors.size() == 1) {
+                throw new VoldemortApplicationException(errors.get(0));
+            } else {
+                throw new VoldemortApplicationException(" Number of Validation failures "
+                                                        + errors.size() + ". Details : "
+                                                        + Arrays.toString(errors.toArray()));
```
gnb: It might be a whole lot easier to just use `logger.error()` here.

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -259,6 +259,10 @@
     public static final String REPAIRJOB_MAX_KEYS_SCANNED_PER_SEC = "repairjob.max.keys.scanned.per.sec";
     public static final String PRUNEJOB_MAX_KEYS_SCANNED_PER_SEC = "prunejob.max.keys.scanned.per.sec";
     public static final String SLOP_PURGEJOB_MAX_KEYS_SCANNED_PER_SEC = "slop.purgejob.max.keys.scanned.per.sec";
+    public static final String ENABLE_NODE_ID_DETECTION = "enable.node.id.detection";
```
gnb: I think only the first two of these are useful, the last 2 should be removed.

src/java/voldemort/server/Host.java
arunthirupathi: Executing hte code on my linux desktop one gives getHostName give machinename. getCanonicalHostName gives machinename.domain.extention

On our production clusters, both return the same though.

src/java/voldemort/server/HostMatcher.java
arunthirupathi: For our production clusters we can configure it to HOSTNAME, but there is a subtle difference between them.

src/java/voldemort/server/PosixHostMatcher.java
arunthirupathi: In windows host names are case insensitive. There are few community members running on windows.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -96,37 +95,66 @@
...
+
+    public static int getNodeId(VoldemortConfig config, Cluster cluster) {
+        int configNodeId = config.getNodeId();
+        if(configNodeId >= 0) {
+            return configNodeId;
+        }
+        if(!config.isEnableNodeIdDetection()) {
+            // Node Id is missing and auto detection is disabled, error out.
+            throw new VoldemortException(VoldemortConfig.NODE_ID
+                                         + " is a required proeprty of the Voldmeort Server");
```
gnb: s/Voldmeort/Voldemort/

src/java/voldemort/server/VoldemortConfig.java
arunthirupathi: node.id.implementation is useful for Testing. node.id.host.types are useful when you only want host name or FQDN matching.

src/java/voldemort/common/nio/SelectorManagerWorker.java
```diff
@@ -122,12 +120,11 @@ else if(!selectionKey.isValid())
             reportException(e);
             close();
         } catch(IOException e) {
-            logger.info("Connection reset from " + socketChannel.socket() + " with message - "
-                        + e.getMessage());
+            logger.info("Connection reset from " + getDebugInfo() + " with message - " + e.getMessage());
```
gnb: This message is misleading.  The entire request handling logic is run in overloads of the `read()` method, so connection reset i.e. the Unix `ECONNRESET` error, is only one of many possible reasons `IOException` might escape out here.  The message should say something like "IO exception from " ...

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
```diff
@@ -80,6 +80,12 @@ public ClientRequestExecutor(Selector selector,
 
     }
 
+    @Override
+    protected String getDebugInfo() {
+        return "Destination: " + String.valueOf(socketDesination) + " . Socket: "
```
gnb: I love this technique for printing more useful stuff in error messages, but the period "." is going to look odd when formatted in some of the messages in SelectorManagerWorker.java

test/unit/voldemort/client/ClientTrafficVerifier.java
```diff
@@ -142,8 +142,7 @@ public void run() {
                 // as it is irrelevant for the refactoring I am doing.
 
             } catch(Exception e) {
-                logger.info("CLIENT EXCEPTION FAILURE on key [" + k + "]" + e.toString());
-                e.printStackTrace();
```
gnb: Die, `printStackTrace()`, die!

test/unit/voldemort/server/VoldemortServerTest.java
```diff
@@ -74,16 +74,24 @@ private VoldemortServer getVoldemortServer(Properties properties) throws IOExcep
     }
 
     @Test
-    public void testJCEProvider() throws IOException {
+    public void testSSLProviders() throws IOException {
+        // The SSL providers changes the state of the JVM
+        // if the methods are run in parallel or the ordering
+        // changes the test fails. TestBouncyCastle is the
+        // only test to enable this SSL. If other tests introduce
+        // them the test could fails as well.
```
gnb: Well caught sir!

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
arunthirupathi: Any suggestions, ignore it or replace it with comma ?

src/java/voldemort/common/nio/SelectorManagerWorker.java
arunthirupathi: Agreed. I never thought about changing this error message, though I have been biten by this in the past. I will change it to IOException from

test/unit/voldemort/server/VoldemortServerTest.java
arunthirupathi: This is something I regressed(It is in the right way in the master). The current one has a single method for it, I tried to improve, but regressed it. This time, I am adding a comment to prevent further regressions :) https://github.com/voldemort/voldemort/blob/master/test/unit/voldemort/server/VoldemortServerTest.java#L28

src/java/voldemort/server/Host.java
gnb: Yeah, so for all the cases that matter, you don't need both.

src/java/voldemort/server/PosixHostMatcher.java
gnb: FQDNs are always case insensitive.  This is a property of [the DNS architecture](https://tools.ietf.org/html/rfc1035) and doesn't depend on the OS of any host.

> Note that while upper and lower case letters are allowed in domain
> names, no significance is attached to the case.  That is, two names with
> the same spelling but different case are to be treated as if identical.

src/java/voldemort/server/VoldemortConfig.java
gnb: > node.id.implementation is useful for Testing.
I suspect it isn't, because you haven't used it that way.

> node.id.host.types are useful when you only want host name or FQDN matching.
in other words, not useful at all.

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
gnb: Sure, commas work.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1388,19 +1388,24 @@ public void operate() {
...
                                   null);
-                metadataStore.put(key,
+                metadataStore.put(keyBytes,
                                   versionedValue,
                                   null);
+                
+                if(MetadataStore.CLUSTER_KEY.equals(keyString)) {
+                    server.handleClusterUpdate();
+                } else if(MetadataStore.NODE_ID_KEY.endsWith(keyString)) {
+                    server.refreshNodeIdFromMetadata();
```
gnb: The ability to set the in-memory node id *without* persisting it, is just downright dangerous.  It will lead to a running service which will fail the next time it's restarted.  Wings fall off button

gnb:
Otherwise, LGTM.  Very pleased to have this @arunthirupathi :smile

src/java/voldemort/server/Host.java
arunthirupathi: Supporting only FQDN and the config is removed.

src/java/voldemort/server/HostMatcher.java
arunthirupathi: Removed the config.

src/java/voldemort/server/PosixHostMatcher.java
arunthirupathi: Removed the PosixHostMatcher class.

src/java/voldemort/server/VoldemortConfig.java
arunthirupathi: Tests directly update the inMemory representation. Have removed the config and this can only be set programatically now. Will add the config if required later.

src/java/voldemort/server/VoldemortServer.java
arunthirupathi: Done.

src/java/voldemort/common/nio/SelectorManagerWorker.java
arunthirupathi: Changed it.

src/java/voldemort/store/socket/clientrequest/ClientRequestExecutor.java
arunthirupathi: Done.

src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
arunthirupathi: I believe it is less dangerous to use this feature with auto detect node id. There is reliable test written to make sure that this works for the client. Some admin operation may be at issues. I fixed more bugs for this code path but there could be some minor issues left.

arunthirupathi:
Addressed the comments, re-ran the unit tests and merged it.

PR #429, opened 2016-07-15
Run store/schema verification in parallel to reduce BnP job running time

gaojieliu:
Run store/schema verification in parallel to reduce BnP job running time.

@arunthirupathi @FelixGV

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
@@ -141,8 +141,10 @@ public class VoldemortBuildAndPushJob extends AbstractJob {
     public final static String MIN_NUMBER_OF_RECORDS = "min.number.of.records";
     public final static String REDUCER_OUTPUT_COMPRESS_CODEC = "reducer.output.compress.codec";
     public final static String REDUCER_OUTPUT_COMPRESS = "reducer.output.compress";
+    public final static String STORE_VERIFICATION_MAX_THREAD_NUM = "store.verification.max.thread.num";
     // default
     private final static String RECOMMENDED_FETCHER_PROTOCOL = "webhdfs";
+    private final int DEFAULT_THREAD_NUM_FOR_STORE_VERIFICATION = 20;
 
     // CONFIG VALUES (and other immutable state)
     private final Props props;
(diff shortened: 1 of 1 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
```diff
@@ -309,6 +312,8 @@ public VoldemortBuildAndPushJob(String name, azkaban.utils.Props azkabanProps) {
         }
         this.executorService = Executors.newFixedThreadPool(requiredNumberOfThreads);
 
+        this.storeVerificationExecutorService = Executors.newFixedThreadPool(THREAD_NUM_FOR_STORE_VERIFICATION);
```
arunthirupathi: In my opinion, there should be a method called preProcess before runBuildStore, which validates the store.

This method creates the thread pool and in the finally block, closes it. In the future, if the thread pool is required at other places, we can extend the scope, but to start with we can make it as a limited scope.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
```diff
@@ -168,6 +168,9 @@
     private final List<Closeable> closeables = Lists.newArrayList();
     private final ExecutorService executorService;
     private final boolean enableStoreCreation;
+    // Executor to do store/schema verification in parallel
+    private final int THREAD_NUM_FOR_STORE_VERIFICATION = 20;
```
arunthirupathi: Can this be made, configurable from the job.properties ?

If Hadoop has limtiation on the thread handles, this way we can work around for smaller time with config parameter.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
gaojieliu: Actually, I want to reuse the thread pool, which can be used by multiple cluster verification.
Otherwise, I need to create a new pool for each cluster verification.

Since there is no requirement to add more pre-process steps right now, the current logic should be acceptable since it is just store verification.
If there are more steps later on, we can wrap them up to a separate function.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
gaojieliu: Sure, I can definitely expose it through job properties.

gaojieliu:
@arunthirupathi 
I exposed one new prop: store.verification.max.thread.num, and if it is less than 1, the job will use sequential store verification logic.

FelixGV:
LGTM.

arunthirupathi:
Looks good to me.

PR #430, opened 2016-07-19
Concurrent insufficient exception

arunthirupathi:
Please review only this (InsufficientOperationalNodes concurrent exception) https://github.com/voldemort/voldemort/commit/7e86b62038d783b441026cdb0b3d45f2d86b93e1

Other 2 commits are test only commits and you can skip it if required.

### DIFF
--- a/src/java/voldemort/store/InsufficientOperationalNodesException.java
+++ b/src/java/voldemort/store/InsufficientOperationalNodesException.java
@@ -17,8 +17,6 @@
 package voldemort.store;
 
 import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 
 import org.apache.log4j.Logger;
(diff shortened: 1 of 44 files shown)


src/java/voldemort/store/InsufficientOperationalNodesException.java
```diff
@@ -35,56 +33,45 @@
...
-                                                 Collection<? extends Throwable> failures) {
+                                                 List<? extends Throwable> failures) {
         this(message + " Original replication set :" + stripNodeIds(replicationSet)
                      + " Known failed nodes before operation :"
                      + stripNodeIds(difference(replicationSet, preferenceList))
                      + " Estimated live nodes in preference list :" + stripNodeIds(preferenceList)
                      + " New failed nodes during operation :"
                      + stripNodeIds(difference(failedList, replicationSet)),
-             failures.size() > 0 ? failures.iterator().next() : null);
+             failures.size() > 0 ? failures.get(0) : null);
```
FelixGV: Why do we even provide a constructor that accepts a list of failures if we're going to cherry pick the first one and discard the rest?

I see that you're deleting the unused `getCauses()` function. That's fine if we don't use it, but if we're going to clean up this class, then I think we should also clean up the constructor. It's a bit misleading the way it is now...

arunthirupathi: More the code change, more possibility of regression. It could be written much better, but I don't want to fix it, unless there is a problem.

arunthirupathi:
@SidW  I believe both of them will do the job fine here and this code path is only triggered when there is failure, which will be less likely. The other code paths are already using copyOnWriteArrayList, so I would like to use the same to follow the same convention.

FelixGV:
Nitpick on the `verifyOrAddStore` time logging: could we tweak the formatting so that the various pieces of info you log are well separated? Maybe:

`INFO verifyOrAddStore() BootStrapUrls: [tcp://localhost:51903], Store: 'abc-xyz-read-only', Verification Time: 10 ms, Creation Time: 35 ms [main]`

FelixGV:
LGTM.

arunthirupathi:
@FelixGV  Fixed the format of the comment, will push it to the master.

PR #432, opened 2016-07-28
RO Store Create floods the Log with error messages

arunthirupathi:
Creating a RO store queries for an existing store, which
fails with StoreNotFoundException. This exception is logged with call
stack, which floods the logs on every store creation.

This may trick the alerting system into treating this as error.
Not logging a call stack and reducing the log to info, when such
exceptions are logged.

### DIFF
--- a/src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
+++ b/src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
@@ -67,6 +67,7 @@ import voldemort.store.StorageEngine;
 import voldemort.store.StoreCapabilityType;
 import voldemort.store.StoreDefinition;
 import voldemort.store.StoreDefinitionBuilder;
+import voldemort.store.StoreNotFoundException;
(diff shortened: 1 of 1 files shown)


src/java/voldemort/server/protocol/admin/AdminServiceRequestHandler.java
```diff
@@ -1505,7 +1506,13 @@ public void operate() {
 
         } catch(VoldemortException e) {
             response.setError(ProtoUtils.encodeError(errorCodeMapper, e));
-            logger.error("handleGetMetadata failed for request(" + request.toString() + ")", e);
+            String errorMessage = "handleGetMetadata failed for request(" + request.toString() + ")";
+            
+            if(e instanceof StoreNotFoundException) {
```
FelixGV: Nitpick: Two catch blocks with two lines each, only one of which is duplicated, seems cleaner to me than trying to DRY this code by using `instanceof`.

FelixGV:
Left a minor nitpick. Feel free to ignore if you disagree. LGTM.

PR #434, opened 2016-08-04
Provide chunk size suggestion for BnP jobs with chunk overflow except…

mattwisein:
…ions and fix num chunks algorith to round up

I'm looking for feedback on this change, I'm still in the process of testing it, so it is not ready to merge.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
@@ -55,6 +55,27 @@ public class HadoopStoreWriter
 
     private static final Logger logger = Logger.getLogger(HadoopStoreWriter.class);
 
+    /**
+     * For each chunk in a partition, Index files contains the integer offset into
+     * the corresponding data file. So it is capped at {@link Integer#MAX_VALUE}.
+     */
+    public static final long MAX_CHUNK_SIZE = Integer.MAX_VALUE;
+
+    /**
+     * The default chunk size tries to leave an error margin so that compressed
+     * data size on hadoop or skewed data sets don't cause chunk overflows
+     */
+    public static final long DEFAULT_CHUNK_SIZE = MAX_CHUNK_SIZE / 2;
+
+    /**
+     * The max number of keys which can collide to the same hash is based on the
+     * fact that we use a short to store the number of keys hashing to this value
+     * in the {@link voldemort.store.readonly.ReadOnlyStorageFormat#READONLY_V2}
+     * format. Therefore, the number of keys which hash to the same value cannot
+     * exceed {@link Short#MAX_VALUE}.
+     */
+    public static final long MAX_HASH_COLLISIONS = Short.MAX_VALUE;
+
     private Set<Integer> chunksHandled = Sets.newHashSet();
 
     private DataOutputStream[] indexFileStream = null;
@@ -289,7 +310,7 @@ public class HadoopStoreWriter
         if(numTuples < 0) {
             // Overflow
             throw new VoldemortException("Found too many collisions: chunk " + chunkId
-                                         + " has exceeded " + Short.MAX_VALUE + " collisions.");
+                                         + " has exceeded " + MAX_HASH_COLLISIONS + " collisions.");
         } else if(numTuples > 1) {
             // Update number of collisions + max keys per collision
             reporter.incrCounter(CollisionCounter.NUM_COLLISIONS, 1);
@@ -327,7 +348,7 @@ public class HadoopStoreWriter
 
         if(this.position[chunkId] < 0)
             throw new VoldemortException("Chunk overflow exception: chunk " + chunkId
-                                         + " has exceeded " + Integer.MAX_VALUE + " bytes.");
+                                         + " has exceeded " + MAX_CHUNK_SIZE + " bytes.");
 
     }
 
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
@@ -61,6 +61,7 @@ import voldemort.store.readonly.ReadOnlyUtils;
 import voldemort.store.readonly.checksum.CheckSum;
 import voldemort.store.readonly.checksum.CheckSum.CheckSumType;
 import voldemort.store.readonly.checksum.CheckSumMetadata;
+import voldemort.store.readonly.disk.HadoopStoreWriter;
 import voldemort.store.readonly.disk.KeyValueWriter;
 import voldemort.store.readonly.mr.azkaban.AbstractHadoopJob;
 import voldemort.store.readonly.mr.azkaban.VoldemortBuildAndPushJob;
@@ -78,8 +79,6 @@ import voldemort.xml.StoreDefinitionsMapper;
 public class HadoopStoreBuilder extends AbstractHadoopJob {
 
     public static final String AVRO_REC_SCHEMA = "avro.rec.schema";
-    public static final long MIN_CHUNK_SIZE = 1L;
-    public static final long MAX_CHUNK_SIZE = (long) (1.9 * 1024 * 1024 * 1024);
     public static final int DEFAULT_BUFFER_SIZE = 64 * 1024;
     public static final short HADOOP_FILE_PERMISSION = 493;
 
@@ -98,7 +97,6 @@ public class HadoopStoreBuilder extends AbstractHadoopJob {
     private final CheckSumType checkSumType;
     private final boolean saveKeys;
     private final boolean reducerPerBucket;
-    private int numChunksOverride;
     private final boolean isAvro;
     private final Long minNumberOfRecords;
     private final boolean buildPrimaryReplicasOnly;
@@ -161,16 +159,6 @@ public class HadoopStoreBuilder extends AbstractHadoopJob {
         this.isAvro = isAvro;
         this.minNumberOfRecords = minNumberOfRecords == null ? 1 : minNumberOfRecords;
         this.buildPrimaryReplicasOnly = buildPrimaryReplicasOnly;
-
-        if(numChunksOverride <= 0) {
-            logger.info("HadoopStoreBuilder constructed with numChunksOverride <= 0, thus relying chunk size.");
-            if(chunkSizeBytes > MAX_CHUNK_SIZE || chunkSizeBytes < MIN_CHUNK_SIZE) {
-                throw new VoldemortException("Invalid chunk size, chunk size must be in the range "
-                        + MIN_CHUNK_SIZE + "..." + MAX_CHUNK_SIZE);
-            }
-        } else {
-            logger.info("HadoopStoreBuilder constructed with numChunksOverride > 0, thus ignoring chunk size.");
-        }
     }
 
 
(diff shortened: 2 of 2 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -409,6 +380,69 @@ public void build() {
...
+     */
+    private void setNumChunksAndNumReduceTasks(long size, JobConf conf) {
+        logger.info("Data size = " + size +
+                ", replication factor = " + storeDef.getReplicationFactor() +
+                ", numNodes = " + cluster.getNumberOfNodes() +
+                ", numPartitions = " + cluster.getNumberOfPartitions() +
+                ", chunk size = " + chunkSizeBytes);
+
+        // Base numbers of chunks and reducers, will get modified according to various settings
+        int numChunks = (int) (size / cluster.getNumberOfPartitions() / chunkSizeBytes) + 1; /* +1 so we round up */
```
mattwisein: Note that the +1 is added on this line.  The refactoring of this method makes it harder to see that it is a change.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -289,14 +237,37 @@ public void build() {
...
+                    long mapOutputBytes = counters.getCounter(Task.Counter.MAP_OUTPUT_BYTES);
+                    int numChunks = conf.getInt(AbstractStoreBuilderConfigurable.NUM_CHUNKS, 1);
+                    long averageNumberOfBytesPerChunk = mapOutputBytes / numChunks / cluster.getNumberOfPartitions();
+                    if (averageNumberOfBytesPerChunk > (HadoopStoreWriter.DEFAULT_CHUNK_SIZE)) {
+                        float chunkSizeBloat = averageNumberOfBytesPerChunk / (float) HadoopStoreWriter.DEFAULT_CHUNK_SIZE;
+                        long suggestedTargetChunkSize = (long) (HadoopStoreWriter.DEFAULT_CHUNK_SIZE / chunkSizeBloat);
+                        logger.error("The number of bytes per chunk may be too high." +
+                            " averageNumberOfBytesPerChunk = " + averageNumberOfBytesPerChunk +
+                            ". Consider setting " + VoldemortBuildAndPushJob.BUILD_CHUNK_SIZE +
+                            "=" + suggestedTargetChunkSize);
```
mattwisein: This block here I changed around from Felix's previous commit.  It now calculates an explicit suggestion and fails instead of re-trying.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
```diff
@@ -55,6 +55,28 @@
 
     private static final Logger logger = Logger.getLogger(HadoopStoreWriter.class);
 
+    /**
+     * The max chunk size is based on the fact that we use an integer for the bytes
```
arunthirupathi: For each chunk in a partition, Index files contains the integer offset into the corresponding data file. So it is capped at Integer.MAX_VALUE

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
```diff
@@ -55,6 +55,28 @@
...
+    public static final long MAX_CHUNK_SIZE = Integer.MAX_VALUE;
+
+    /**
+     * The default chunk size tries to leave an error margin so that compressed
+     * data size on hadoop or skewed data sets don't cause chunk overflows
+     */
+    public static final long DEFAULT_CHUNK_SIZE = MAX_CHUNK_SIZE / 2;
+
+    /**
+     * The max number of keys which can collide to the same hash is based on the
```
arunthirupathi: Can the constant be named MAX_HASH_COLLISIONS ?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -289,14 +237,37 @@ public void build() {
...
+            RunningJob runningJob = jc.submitJob(conf);
+            Counters counters = runningJob.getCounters();
+
+            try {
+                if (!jc.monitorAndPrintJob(conf, runningJob)) {
+                    // For some datasets, the number of chunks that we calculated is inadequate.
+                    // Here, we try to identify if this is the case.
+                    long mapOutputBytes = counters.getCounter(Task.Counter.MAP_OUTPUT_BYTES);
+                    int numChunks = conf.getInt(AbstractStoreBuilderConfigurable.NUM_CHUNKS, 1);
+                    long averageNumberOfBytesPerChunk = mapOutputBytes / numChunks / cluster.getNumberOfPartitions();
```
arunthirupathi: does this have same ceil issue that we are fixing ? but since this is bytes, probably not that important.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -289,14 +237,37 @@ public void build() {
...
+
+            JobClient jc = new JobClient(conf);
+            RunningJob runningJob = jc.submitJob(conf);
+            Counters counters = runningJob.getCounters();
+
+            try {
+                if (!jc.monitorAndPrintJob(conf, runningJob)) {
+                    // For some datasets, the number of chunks that we calculated is inadequate.
+                    // Here, we try to identify if this is the case.
+                    long mapOutputBytes = counters.getCounter(Task.Counter.MAP_OUTPUT_BYTES);
```
arunthirupathi: This is definitely a better heuristic, but still may not be right if there is a data skew. This is sum of all partition data sizes and with MD5 the distribution is supposed to be even. But if there is a data skew, the counter that is best is max partition size .  

But even then, our chunking is based on the hash, so it is impossible to come up with an accurate number.

FelixGV: You are right that this does not solve the issue of data skew.

In practice, I don't think I've seen a case where data skew was actually the issue that triggered a chunk overflow. It's theoretically possible, but the fact that this category of edge cases can only happen in big data sets (those on the order of: partition count * chunk size) makes it unlikely that there would be significant skew in such big data set.

That being said, I don't think the MR counters can give us per mapper stats out of the box. We could record it ourselves in custom counters, but I don't think it's worth it.

I think we can simply apply a safety margin (like 0.75 or even 0.5) on top of the properly measured chunk size, and we'd be safe enough.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -289,14 +237,37 @@ public void build() {
...
+                    int numChunks = conf.getInt(AbstractStoreBuilderConfigurable.NUM_CHUNKS, 1);
+                    long averageNumberOfBytesPerChunk = mapOutputBytes / numChunks / cluster.getNumberOfPartitions();
+                    if (averageNumberOfBytesPerChunk > (HadoopStoreWriter.DEFAULT_CHUNK_SIZE)) {
+                        float chunkSizeBloat = averageNumberOfBytesPerChunk / (float) HadoopStoreWriter.DEFAULT_CHUNK_SIZE;
```
FelixGV: I imagine you meant `Float` rather than `Bloat`?

Also, aren't you calculating a number of chunks here, rather than a chunk size?

I suggest renaming to `properNumChunksFloat` or something like that.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -289,14 +237,37 @@ public void build() {
...
+                        long suggestedTargetChunkSize = (long) (HadoopStoreWriter.DEFAULT_CHUNK_SIZE / chunkSizeBloat);
```
FelixGV: As I mentioned in another comment above, I'd like us to apply a safe margin here. Can you define a constant such as `float CHUNK_SIZE_SAFETY_MARGIN = 0.75` and multiply it to your number here?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
```diff
@@ -409,6 +380,69 @@ public void build() {
...
+     * On the other hand, if the size is that of the mappers' output, then it should be possible
+     * to predict proper settings. There is still another edge case, however, which we cannot
+     * avoid completely: if the data is heavily skewed towards a hot partition, then it may not
+     * be possible at all to tweak these settings correctly.
+     *
+     * @param size of the dataset
+     * @param conf MapReduce {@link JobConf} for which to set properties.
+     * @return
+     */
+    private void setNumChunksAndNumReduceTasks(long size, JobConf conf) {
```
FelixGV: Since we are not retrying the job automatically in this modified PR, unlike my original PR, I don't think it's worth having this logic in a separate function. It makes the flow more confusing without actually reducing code size, since the function is now called only once. Furthermore, the java doc is a bit misleading, since there is now never any case where we call this function with the mapper's output as the `size` parameter.

For these reasons, I'd recommend leaving this code where it was originally, rather than moving it into a function.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
mattwisein: I'm not sure what you're suggesting here.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
mattwisein: Yeah, just bytes. Here we divide by the num chunks that was generated earlier so its dividing a big number by a small number to get a big number.  Rounding from integer division is not an issue.

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/disk/HadoopStoreWriter.java
arunthirupathi: I was suggesting to re-word the comments :)

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
mattwisein: The default chunk size is currently the max chunk size / 2, which gives us a factor of 2 safety margin.   (Which is actually a safety margin now that the integer division is resolved)

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
mattwisein: No, I mean Bloat.  This is a ratio, an estimate of how much the chucks are getting bloated when they are decompressed.  Then we use that ratio to suggest a chunk size setting that when applied to the compressed HDFS data, will yield decompressed chunks of `DEFAULT_CHUNK_SIZE` (or 1gb)

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/HadoopStoreBuilder.java
mattwisein: the default chunk size takes this into account.

mattwisein:
Tested!!

PR #440, opened 2016-08-17
Data Cleanup job Does not run on system stores

arunthirupathi:
1) Client registry System store is a in-memory store and supposed to be cleaned up after 7 days.
Last change to the DataCleanupJob made the system stores fail with the missing store exception.

Clients re-use the same client id, so unless lots of clients become
dead and removed, this will not cause a leak on the server resources. The effect is negligible.

Now the DataCleanupJob checks for both system stores and normal stores for a store definition.

2) If the store retention days is modified to zero, then the store will
delete all the records. But if the store is started with 0 retention days
it means the data retention is not enabled. Fixed the discrepancy.

### DIFF
--- a/src/java/voldemort/store/metadata/MetadataStore.java
+++ b/src/java/voldemort/store/metadata/MetadataStore.java
@@ -1168,6 +1168,20 @@ public class MetadataStore extends AbstractStorageEngine<ByteArray, byte[], byte
         }
     }
 
+    public static StoreDefinition getStoreDef(String storeName, MetadataStore metadataStore) {
+        StoreDefinition def = null;
+        if(SystemStoreConstants.isSystemStore(storeName)) {
+            def = SystemStoreConstants.getSystemStoreDef(storeName);
+        } else {
+            def = metadataStore.getStoreDef(storeName);
+        }
+
+        if(def == null) {
+            throw new StoreNotFoundException("Store" + storeName + " does not exist");
+        }
+        return def;
+    }
+
     /**
      * Initializes the metadataCache for MetadataStore
      */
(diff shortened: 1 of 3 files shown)


src/java/voldemort/store/metadata/MetadataStore.java
```diff
@@ -1168,6 +1168,16 @@ public void initNodeId(int nodeId) {
         }
     }
 
+    public static StoreDefinition getStoreDef(String store, MetadataStore metadataStore) {
+        StoreDefinition def = null;
+        if(SystemStoreConstants.isSystemStore(store)) {
```
gaojieliu: This function could return null if store is treated as system store, but the definition doesn't exist, since 'SystemStoreConstants.isSystemStore' only checks the store name prefix.
Normally, it won't happen, but I think it is better to throw VeniceException when it is null to keep the logic consistent.

arunthirupathi: This function is lifted from here. https://github.com/voldemort/voldemort/blob/master/src/java/voldemort/server/protocol/admin/FetchStreamRequestHandler.java#L143
I did not make the code use the same method as I don't want to test the Stream request handling part.

DataCleanupJob is only scheduled for existing stores, so null is never a possibility.  But since this can be used in the future, I will add the null check.

gaojieliu:
For the following scenario:
1. The store retention period is initialized to be 0 when started;
2. Later on, admin changes the store retention period to be non-zero;
Then the clean up job won't start for the later change, is it expected?
If we want to support this scenario, maybe we should run clean up job for each store no matter the retention period is 0 or not.

arunthirupathi:
Yep, that is expected. When SRE adds a retention, it requires server bounce. 

The wording is not clear in this commit message, but this is what I meant.
https://github.com/voldemort/voldemort/commit/71280330a8ecb6a60409f0a2594445ef568da150

gaojieliu:
Got it, thanks @arunthirupathi

arunthirupathi:
@gaojieliu  the code review is updated. Please let me know if you have further comments.

gaojieliu:
LGTM!

arunthirupathi:
Thanks @gaojieliu the tests are running will merge this in, when the tests complete.

PR #441, opened 2016-08-17
vadmin.sh stream support for system stores

arunthirupathi:
The commands
bin/vadmin.sh stream fetch-entries
bin/vadmin.sh stream fetch-keys

does not work on System stores like voldsys$_client_registry

There is a client side check for valid stores, which only is
validating the user stores. Added a check to include the system
stores as well.


gaojieliu:
I saw there is one more command: stream mirror, which is using 'validateUserStoreNamesOnNode' as well, and if this command is planned for system stores, then the code change looks good.

arunthirupathi:
The truncate partition also uses the same method. But I tried using the newer method but server does not have the support, so decided to fix only the part that works. Will merge this in.

PR #443, opened 2016-08-25
Tweaked the semantics of the "readonly.modify.port" server configuration.

FelixGV:
Tweaked the semantics of the "readonly.modify.port" server configuration.

Previously, -1 (the default) meant that the port will not be modified.

Now, -1 means that the port will be removed from the fetch URI, while
null (the new default) means that the port will not be modified.

Therefore, the default behavior is still the same.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/VoldemortUtils.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/VoldemortUtils.java
@@ -144,8 +144,8 @@ public class VoldemortUtils {
         return new StoreDefinitionsMapper().readStore(new StringReader(xml));
     }
 
-    public static String modifyURL(String originalUrl, String newProtocol, int newPort) {
-        if (newProtocol == null || newProtocol.isEmpty() || newPort < 0) {
+    public static String modifyURL(String originalUrl, String newProtocol, Integer newPort) {
+        if (newProtocol == null || newProtocol.isEmpty() || newPort == null) {
             return originalUrl;
         }
 
--- a/src/java/voldemort/server/VoldemortConfig.java
+++ b/src/java/voldemort/server/VoldemortConfig.java
@@ -349,7 +349,7 @@ public class VoldemortConfig implements Serializable {
         // To enable block-level compression over the wire for Read-Only fetches, set this property to "GZIP"
         defaultConfig.put(READONLY_COMPRESSION_CODEC, "NO_CODEC");
         defaultConfig.put(READONLY_MODIFY_PROTOCOL, "");
-        defaultConfig.put(READONLY_MODIFY_PORT, -1);
+        defaultConfig.put(READONLY_MODIFY_PORT, (Integer) null);
         defaultConfig.put(USE_BOUNCYCASTLE_FOR_SSL, false);
         defaultConfig.put(READONLY_BUILD_PRIMARY_REPLICAS_ONLY, true);
 
@@ -595,7 +595,7 @@ public class VoldemortConfig implements Serializable {
     private long readOnlyLoginIntervalMs;
     private long defaultStorageSpaceQuotaInKB;
     private String modifiedProtocol;
-    private int modifiedPort;
+    private Integer modifiedPort;
     private boolean bouncyCastleEnabled;
     private boolean readOnlyBuildPrimaryReplicasOnly;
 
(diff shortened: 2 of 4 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/utils/VoldemortUtils.java
```diff
@@ -144,8 +144,8 @@ public static StoreDefinition getStoreDef(String xml) {
         return new StoreDefinitionsMapper().readStore(new StringReader(xml));
     }
 
-    public static String modifyURL(String originalUrl, String newProtocol, int newPort) {
-        if (newProtocol == null || newProtocol.isEmpty() || newPort < 0) {
+    public static String modifyURL(String originalUrl, String newProtocol, Integer newPort) {
+        if (newProtocol == null || newProtocol.isEmpty() || newPort == null) {
```
mattwisein: Do we still want to check if newPort<0 ?  I don't fully grasp the point of your change, but there might be a regression if some code is still expecting to get away with setting a port as -1.

src/java/voldemort/server/VoldemortConfig.java
```diff
@@ -880,7 +880,7 @@ private void initializeStateFromProps() throws UndefinedPropertyException {
         this.readOnlyMaxValueBufferAllocationSize = this.allProps.getInt(READONLY_MAX_VALUE_BUFFER_ALLOCATION_SIZE);
         this.readOnlyCompressionCodec = this.allProps.getString(READONLY_COMPRESSION_CODEC);
         this.modifiedProtocol = this.allProps.getString(READONLY_MODIFY_PROTOCOL);
-        this.modifiedPort = this.allProps.getInt(READONLY_MODIFY_PORT);
+        this.modifiedPort = this.allProps.getNullableInt(READONLY_MODIFY_PORT);
```
mattwisein: Does `getNullableInt` return an int or an Integer?  If the latter, we could rename it to `getNullableInteger` just for clarity sake.

src/java/voldemort/utils/Props.java
```diff
@@ -128,7 +128,11 @@ public String put(String key, String value) {
     }
 
     public String put(String key, Integer value) {
-        return props.put(key, value.toString());
+        if (null == value) {
```
arunthirupathi: This is a change in the common path, have you examined all the callers to see if someone is not dependent on this behavior.

Also it seems hackish, when setting a value to null, removes the pre-existent value on a base method. It should go in a different method, like reset or remove.

arunthirupathi:
What is the scenario, driving this change ?

The change is not backward compatible for users, who have the port specified. It is OK, if you think that is a bad behavior.

Also a base method, when you inserting a null, deleting the property is very hacky. It could cause issues down the line. If you want it only for a specific case, please introduce a new method.

src/java/voldemort/server/VoldemortConfig.java
FelixGV: It's implied that it returns an `Integer`, since a primitive `int` cannot be null. Anyhow, I've renamed the function to `getNullableInteger` as you suggest.

src/java/voldemort/utils/Props.java
FelixGV: This is inline with the already existing `getNullableInt()` function which returns null if the key is not contained.

The previous code did not allow passing in a null value at all, since it results in a NPE at the time the `put` is called. As far as I can tell, there is no code that depends on receiving a NPE when passing in a null.

That being said, I agree that this isn't super clean. Perhaps I should create a new config rather than overloading the meaning of the existing config.

FelixGV:
@arunthirupathi: The motivation behind this change is to introduce the ability to remove the port completely from a Hadoop fetch URI. We have a new (internal) Hadoop library coming up soon which has no port as part of its scheme and I want to be able to support it...

The change is compatible for people who explicitly defined the config setting as any positive integer. It is also compatible for people who did not define the config setting. The semantic is changed for people who explicitly defined the config setting to -1.

Like I said in another comment, I agree that this may not be a clean change. Perhaps I should create a new config, such as `READONLY_OMIT_PORT` (boolean, defautl false), rather than overloading the meaning of the existing config...

FelixGV:
FYI: I've created the new PR #444 in order to introduce a new config, rather than altering the semantics of the existing config.

PR #444, opened 2016-08-29
Introduced new boolean "readonly.omit.port" server configuration.

FelixGV:
When set to true, the port will be removed from the fetch URI. In
this case, the already-existing "readonly.modify.port" setting is
ignored.

When set to false (which is the default), then the port will be
left as part of the fetch URI (according to the already-existing
"readonly.modify.port" setting).


mattwisein:
LGTM

FelixGV:
Thanks @mattwisein 

This is merged.

PR #447, opened 2016-09-12
The BnP job should be resilient to colo failures, but this has regressed

FelixGV:
Adding a safe guard to bring back resilience, in the BnP constructor.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
@@ -47,6 +47,7 @@ import org.apache.hadoop.mapred.JobConf;
 import org.apache.log4j.Logger;
 
 import voldemort.VoldemortException;
+import voldemort.client.BootstrapFailureException;
 import voldemort.client.ClientConfig;
 import voldemort.client.protocol.admin.AdminClient;
 import voldemort.client.protocol.admin.AdminClientConfig;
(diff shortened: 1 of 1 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
```diff
@@ -212,10 +214,18 @@ public VoldemortBuildAndPushJob(String name, azkaban.utils.Props azkabanProps) {
...
-                this.clusterURLs.add(url);
-                AdminClient adminClient = createAdminClient(url, fetchAllStoresXml);
-                this.adminClientPerCluster.put(url, adminClient);
-                this.closeables.add(adminClient);
+                try {
+                    AdminClient adminClient = createAdminClient(url, fetchAllStoresXml);
+                    this.clusterURLs.add(url);
+                    this.adminClientPerCluster.put(url, adminClient);
+                    this.closeables.add(adminClient);
+                } catch (Exception e) {
```
gaojieliu: If we simply skip the unreachable colo here, the user may get confusion since the unreachable colo may not have the latest data if some colo is temporarily not available from push job, but the push job does succeed.
The Azkaban logs might be too long for user to understand the status of each colo every time.

FelixGV: Yeah, but that was the default behavior up to some time ago. I'm not sure when we introduced this change in semantics (I could try to find out, if you're curious...).

Are you suggesting that we should fail the job at the end of the `run()` if some cluster was unreachable during initialization?

gaojieliu: Yes.
From user's perspective, successful job means that the data is ready in every colo, and it can be easily inferred through job status.
Otherwise, it might too hard for them to check the job logs every time.
Also it might not be easy for us to track job status since normally we only check whether the job fails or not.

FelixGV: I've altered the `catch` logic so that we keep track of the colo(s) that failed. This will cause the job to fail with a proper error message after it is done pushing to the healthy colos.

gaojieliu:
LGTM!

FelixGV:
Thanks a lot @gaojieliu. I've merged this in.

PR #450, opened 2016-09-26
Fix Voldemort URLs

mattwisein:
Replaced the following instances with http://www.project-voldemort.com since urls don't resolve from some places

$ grep -r 'http://project-voldemort' .
./clients/python/setup.py:      url='http://project-voldemort.com',
./NOTES:For the most up-to-date information see http://project-voldemort.com
./contrib/collections/src/java/voldemort/collections/VStack.java: *        voldemort JSON formats: http://project-voldemort.com/design.php

### DIFF
--- a/NOTES
+++ b/NOTES
@@ -1,6 +1,6 @@
 Getting Started
 
-For the most up-to-date information see http://project-voldemort.com
+For the most up-to-date information see http://www.project-voldemort.com
 
 ## checkout and build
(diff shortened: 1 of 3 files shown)


FelixGV:
LGTM! Thanks Matt!

benmccann:
That's more of a work-around than a fix. The problem still persists in numerous places (including typing the URL manually, on internal wikis, etc.). Really the root problem of project-voldemort.com not resolving should be fixed

PR #452, opened 2016-10-05
postgres storage engine

code-stash:


FelixGV:
Interesting code drop (:

Are you using this PG storage engine code?

-F

code-stash:
Not in any production env. The pluggable architecture was so intuitive, could not resist coming up with this PG storage engine code. Need to do more testing/ code improvements before sending the actual pull request

PR #453, opened 2016-10-19
Adhere target/source java compatibility for contrib projects

bitti:
We are using Java 1.7 on our hadoop cluster, but locally I use JDK 1.8
for building, so contrib got build with 1.8 by default because its
source and target version is not specified explicitly. Consequentially
the resulting jar didn't work in our cluster environment. This
reconfiguration fixes this.

But since the contrib projects don't build anymore with 1.6 source
compatibility I bumped up the configured javac.version to 1.7, even
though two other options are available:
1. Fix contrib to make it 1.6 compatible again
2. Specify different versions for the main and contrib projects

But since Java 1.6 is EOL since quite a while I suggest that no
Voldemort servers with a Java 1.6 runtime should be running anymore and
it should be save to upgrade and keep the configuration simple.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,3 @@
-
 apply plugin: 'java'
 apply plugin: 'war'
 
@@ -76,7 +75,7 @@ allprojects {
             options.addStringOption('Xdoclint:none', '-quiet')
         }
     }
-    
+
     tasks.withType(Javadoc) {
         enabled = javadocEnabled
     }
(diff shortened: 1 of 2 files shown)


FelixGV:
I am ok with bumping all of Voldemort to 1.7.

If no one else raises any concern with this in the next few days, I'll merge it in.

FelixGV:
Sorry for the wait, I've merged this in.

We are currently running on JDK 8 in production (without the new G1 GC). We've been running that for a while without issue, but I'm hesitant to push it higher than 7 for now.

PR #455, opened 2016-10-24
Fix detection for fetcher protocol warning

bitti:
Fetcher protocol warning was shown if the recommended protocol is
specified explicitly, which is unnecessary confusing

I'm not sure what it takes for a PR to be merged, therefore I put out this very minimal PR.


FelixGV:
Oh, yeah, of course. Duh. Sorry for the wait, this is merged in.

PR #458, opened 2016-11-04
Made admin connection/socket timeout configurable in BnP.

FelixGV:
Also changed the default of these configs to 180 seconds.

This fixes the following problem: when a node is unreachable and
completely shut down, requests to it will time out, which takes
60 seconds. When BnP notices this, it will reach one of the live
nodes in the cluster and ask it to deal with the failure. The
live node will try to talk to the dead node, which will also
take 60 seconds to time out. By the time the live node decides
that the dead node is unreachable, and responds to the BnP job,
the BnP job will have already timed out. Then, the BnP job will
think that the HandleFailedFetchRequest could not complete
successfully (even though it did in fact complete successfully)
and BnP HA will be aborted.

The solution is that the BnP job's timeout must be greater than
the server's timeout.

This was not an issue before when we had insanely long time outs,
but those time outs have been reduced considerably in commit
34debd34c5896b6c2a01b1012e89dd1a3a0a0242. This is likely when we
regressed on the handling of this failure mode.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
@@ -145,6 +145,8 @@ public class VoldemortBuildAndPushJob extends AbstractJob {
     public final static String REDUCER_OUTPUT_COMPRESS_CODEC = "reducer.output.compress.codec";
     public final static String REDUCER_OUTPUT_COMPRESS = "reducer.output.compress";
     public final static String STORE_VERIFICATION_MAX_THREAD_NUM = "store.verification.max.thread.num";
+    public final static String ADMIN_CLIENT_CONNECTION_TIMEOUT_SEC = "admin.client.connection.timeout.sec";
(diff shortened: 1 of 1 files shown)


FelixGV:
@mattwisein : Connection time out is for the initial connection establishment, whereas socket time out is for subsequent reads on that connection. We can make one be shorter than the other if we want, but I want both to be higher than the default server-side connection timeout (60 seconds).

mattwisein:
Do we really want to allow a 3 minute timeout for establishing a connection?  It sounds like the failure mode we're trying to avoid is lots of time for transitive queries on a request, not on connection establishment.  60 seconds (or even 15) seems like plenty of time for establishing a connection, no?

FelixGV:
15 seconds ought to be plenty, but because the server's default timeout for contacting another server is 60 seconds, the BnP job's timeout must be higher than that.

This is not ideal, but I'd rather have default settings that "just work", even if they may be a little slow. People can always tune those configs lower in their own deployments if they want.

FelixGV:
I spoke offline with @mattwisein and we agreed that only the socket time out's default value needs to be increased. My amended commit therefore leaves the connection timeout at 15 seconds, like previously, though both values are now configurable.

PR #459, opened 2016-11-07
Changed the DeleteAllFailedFetchStrategy so that it affects all nodes.

FelixGV:
Changed the DeleteAllFailedFetchStrategy so that it affects all nodes.

Previously, the DeleteAllFailedFetchStrategy would only attempt to
delete data from nodes which succeeded in their fetch. In some failure
modes, this is appropriate, but in other cases, it isn't. In any case,
there is no harm in trying to delete data on all nodes, even those
that failed their fetch. This commit makes it so.


gaojieliu:
This change looks good to me.
I have a question about backup cleanup during swap:
Considering there are some failed fetches (which didn't get deleted in some case), the backup deletion logic can retain some partial-fetched store since Voldemort doesn't store the status for each fetch, will it cause any problem?
Please let me know if my understanding is not correct.

FelixGV:
Definitely, partially fetched stores could theoretically be swapped into place which would definitely cause issues.

FelixGV:
Thanks for the review, this is merged in (:

PR #460, opened 2016-11-08
BnP now kills an async job that it is waiting on if that job times out.

FelixGV:
BnP now kills an async job that it is waiting on if that job times out.

Previously, BnP would just leave the aync job running if it timed out,
which is wasteful, and could cause a subsequent job retry to fail if
there are two fetch jobs running concurrently for the same store.


FelixGV:
Thanks for the review, this is merged in (:

PR #461, opened 2016-11-09
Fix race condition with BnP HA when cluster xml changes during a push

FelixGV:
- Tweaked the AdminClient's currentVersion so that it is not stale.
- BnP now retries fetches when cluster.xml is stale.

(See commit messages for details)

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortSwapJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortSwapJob.java
@@ -119,17 +119,21 @@ public class VoldemortSwapJob extends AbstractJob {
                     dataDir, e);
         }
 
-        // While processing an admin request, HDFSFailedLock could take long time because of multiple HDFS operations,
-        // especially when the name node is in a different data center. So extend timeout to 5 minutes.
+        // It should not be necessary to set the max conn / node so high, but it should not be a big deal either. New
+        // connections will be created as needed, not upfront, so there should be no extra cost associated with the
+        // higher setting. There shouldn't be many parallel requests happening in this use case, but we're going to
+        // leave it as is for now, just to minimize the potential for unforeseen regressions.
         AdminClientConfig adminConfig = new AdminClientConfig().setMaxConnectionsPerNode(cluster.getNumberOfNodes())
                                                                .setMaxBackoffDelayMs(maxBackoffDelayMs)
+        // While processing an admin request, HDFSFailedLock could take long time because of multiple HDFS operations,
+        // especially when the name node is in a different data center. So extend timeout to 5 minutes.
                                                                .setAdminSocketTimeoutSec(60 * 5);
 
         ClientConfig clientConfig = new ClientConfig().setBootstrapUrls(cluster.getBootStrapUrls())
                                                       .setConnectionTimeout(httpTimeoutMs,
                                                                             TimeUnit.MILLISECONDS);
         // Create admin client
-        AdminClient client = new AdminClient(cluster, adminConfig, clientConfig);
+        AdminClient client = new AdminClient(adminConfig, clientConfig);
 
         if (pushVersion == -1L) {
             // Need to retrieve max version
@@ -148,7 +152,6 @@ public class VoldemortSwapJob extends AbstractJob {
         // do the fetch, and if it succeeds, the swap
         info("Initiating fetch of " + storeName + " with dataDir: " + dataDir);
         AdminStoreSwapper swapper = new AdminStoreSwapper(
-                cluster,
                 executor,
                 client,
                 httpTimeoutMs,
--- a/src/java/voldemort/client/protocol/admin/AdminClient.java
+++ b/src/java/voldemort/client/protocol/admin/AdminClient.java
@@ -171,9 +171,10 @@ public class AdminClient implements Closeable {
     private final AdminStoreClient adminStoreClient;
     private final NetworkClassLoader networkClassLoader;
     private final AdminClientConfig adminClientConfig;
-    private final long clusterVersion;
+    private final ClientConfig clientConfig;
     private final boolean fetchSingleStore;
     private final String debugInfo;
+    private Long clusterVersion;
     private Cluster currentCluster;
     private SystemStoreClient<String, String> metadataVersionSysStoreClient = null;
     private SystemStoreClientFactory<String, String> systemStoreFactory = null;
@@ -231,14 +232,18 @@ public class AdminClient implements Closeable {
                                                                .getContextClassLoader());
 
         this.adminClientConfig = adminClientConfig;
+        this.clientConfig = clientConfig;
         this.socketPool = helperOps.createSocketPool(adminClientConfig);
         this.adminStoreClient = new AdminStoreClient(clientConfig);
         this.fetchSingleStore = !clientConfig.isFetchAllStoresXmlInBootstrap();
 
-        this.clusterVersion = System.currentTimeMillis();
         if(cluster != null) {
+            // We don't know at which time the Cluster's state was generated, so we can't know the cluster version.
+            this.clusterVersion = null;
             this.currentCluster = cluster;
         } else {
+            // Important: this time must be recorded before we bootstrap from the remote state
+            this.clusterVersion = System.currentTimeMillis();
             this.currentCluster = getClusterFromBootstrapURL(clientConfig);
         }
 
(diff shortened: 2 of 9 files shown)


contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortSwapJob.java
```diff
@@ -121,15 +121,15 @@ public void run() throws Exception {
 
         // While processing an admin request, HDFSFailedLock could take long time because of multiple HDFS operations,
         // especially when the name node is in a different data center. So extend timeout to 5 minutes.
-        AdminClientConfig adminConfig = new AdminClientConfig().setMaxConnectionsPerNode(cluster.getNumberOfNodes())
+        AdminClientConfig adminConfig = new AdminClientConfig().setMaxConnectionsPerNode(cluster.getNumberOfNodes()) // TODO: Figure out why do we need this config?? -FGV
```
arunthirupathi: The number of nodes is a bug. but since Voldemort grows connection on a need basis, I never bothered to fix it. It should be 2 or 3 and never more than that. But it does not hurt as there should not be parallel connections using it.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -315,11 +319,17 @@ public void close() {
      */
     public boolean isClusterModified() {
         try {
-            Properties props = MetadataVersionStoreUtils.getProperties(metadataVersionSysStoreClient);
-            long retrievedVersion = MetadataVersionStoreUtils.getVersion(props,
-                                                                         SystemStoreConstants.CLUSTER_VERSION_KEY);
+            if (null == clusterVersion) {
+                // Then we have no choice to do a full cluster.xml equality check in order to know if the cluster is the same.
+                Cluster remoteCluster = getClusterFromBootstrapURL(clientConfig);
+                return currentCluster.equals(remoteCluster);
```
arunthirupathi: Shouldn't this be not equals ? as the semantics says cluster modified ?

You can cache the time if the check equals to avoid doing the expensive call. But the time needs to be intialized before the cluster is initialized.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -231,14 +232,17 @@ private AdminClient(AdminClientConfig adminClientConfig,
...
         this.adminStoreClient = new AdminStoreClient(clientConfig);
         this.fetchSingleStore = !clientConfig.isFetchAllStoresXmlInBootstrap();
 
-        this.clusterVersion = System.currentTimeMillis();
         if(cluster != null) {
+            // We don't know at which time the Cluster's state was generated, so we can't know the cluster version.
+            this.clusterVersion = null;
             this.currentCluster = cluster;
         } else {
+            this.clusterVersion = System.currentTimeMillis();
```
arunthirupathi: Recommend leaving a comment that time needs to be initalized before calling the bootstrap XML.

src/java/voldemort/client/protocol/admin/AdminClient.java
FelixGV: Oh duh. Of course, you are totally right. And yes, I will also initialize the time before fetching the `remoteCluster`. Good call. Thanks a lot!

src/java/voldemort/client/protocol/admin/AdminClient.java
FelixGV: Ok, will do.

src/java/voldemort/client/protocol/admin/AdminClient.java
```diff
@@ -4426,16 +4436,16 @@ public String fetchStore(int nodeId,
             int asyncId = response.getRequestId();
             try {
                 return rpcOps.waitForCompletion(nodeId, asyncId, timeoutMs, TimeUnit.MILLISECONDS);
-            } catch (VoldemortException ve) {
-                logger.error("Got an exception from rpcOps.waitForCompletion for nodeId " + nodeId + ", asyncId " +
```
arunthirupathi: I would prefer to keep this to log error with nodeId in the logs. Otherwise we might lack debugging information.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -174,14 +164,32 @@ public String call() throws Exception {
...
+                    } catch (VoldemortException e) {
+                      if (ExceptionUtils.recursiveClassEquals(e, ExceptionUtils.BNP_SOFT_ERRORS)) {
+                          if (adminClient.isClusterModified()) {
+                              logger.info("Got a " + e.getClass().getSimpleName() + " from " + node.briefToString()
+                                          + " while trying to fetch store " + storeName + ". It seems like the"
+                                          + " cluster.xml state has changed since this AdminClient was constructed."
+                                          + " Therefore, we will attempt constructing a fresh AdminClient and retrying"
+                                          + " the fetch operation.");
+                              AdminClient freshAdminClient = adminClient.getFreshClient();
+                              return fetch(storeDir, freshAdminClient);
```
arunthirupathi: This seems like it can go into infinite retries. Can the retry be limited?

contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortSwapJob.java
FelixGV: Ok thanks. I've changed my comment to integrate the info you provide above. I'm leaving the setting as is.

src/java/voldemort/client/protocol/admin/AdminClient.java
arunthirupathi: I believe, It logs only on AsyncOperationException and not on base VoldemortException type. Sorry for not being clear.

FelixGV: Ok, sure, I'll add a logging statement for `VoldemortException` as well.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -174,14 +168,52 @@ public String call() throws Exception {
...
+                        if (ExceptionUtils.recursiveClassEquals(ve, ExceptionUtils.BNP_SOFT_ERRORS)) {
+                            String logMessage = "Got a " + ve.getClass().getSimpleName() + " from " + node.briefToString()
+                                                + " while trying to fetch store '" + storeName + "'"
+                                                + " (attempt " + attempt + "/" + MAX_FETCH_ATTEMPTS + ").";
+                            AdminClient possiblyRefreshedAdminClient = adminClient;
+                            if (adminClient.isClusterModified()) {
+                                logMessage += " It seems like the cluster.xml state has changed since this AdminClient"
+                                            + " was constructed. Therefore, we will attempt constructing a fresh"
+                                            + " AdminClient and retrying the fetch operation.";
+                                possiblyRefreshedAdminClient = adminClient.getFreshClient();
```
squarY: Shall we also need to update this.adminClient to the refreshed one? Looks like we create a new one but and pass it in to recursive "fetch"  but this.adminClient is still the old one.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -174,14 +168,52 @@ public String call() throws Exception {
...
+                            } else {
+                                logMessage += " The cluster.xml is up to date. We will retry with the same AdminClient.";
+                            }
+                            logger.info(logMessage);
+                            return fetch(hadoopStoreDirToFetch, possiblyRefreshedAdminClient, ++attempt);
```
squarY: Could we use a loop instead of tail recursion here?

FelixGV: So you have a concern that there could be a stack overflow? Or is it a stylistic preference? I don't think the former is a big issue, given the maximum number of retries?

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
FelixGV: I added a comment to explain this.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
FelixGV: I rewrote this code as a loop.

FelixGV:
I've merged this in. Thanks a lot for the review guys!

PR #462, opened 2016-11-09
Rolling restart during BnP

FelixGV:
Since I've been fixing BnP issues left right and center since last week, and I had to re-read a lot of that code, I figured I might as well implement BnP rolling restart support as well, while I'm at it... So I did just that in the past 2-3 hours.

This commit is based on #461, so it's actually a bit smaller than it looks. This is the `--shortstat` for just rolling restart part: 8 files changed, 136 insertions(+), 17 deletions(-)

In any case, it is a bit finicky and would probably require extensive review and testing before being signed off.

### DIFF
--- a/src/java/voldemort/server/protocol/admin/AsyncOperationService.java
+++ b/src/java/voldemort/server/protocol/admin/AsyncOperationService.java
@@ -84,7 +84,7 @@ public class AsyncOperationService extends AbstractService {
      */
     public synchronized boolean isComplete(int requestId, boolean remove) {
         if (!operations.containsKey(requestId))
-            throw new VoldemortException("No operation with id " + requestId + " found");
+            throw new AsyncOperationNotFoundException(requestId);
 
         if (operations.get(requestId).getStatus().isComplete()) {
             if (logger.isDebugEnabled())
@@ -112,7 +112,7 @@ public class AsyncOperationService extends AbstractService {
         try {
             return getOperationStatus(id).toString();
         } catch(VoldemortException e) {
-            return "No operation with id " + id + " found";
+            return String.format(AsyncOperationNotFoundException.MESSAGE, id);
         }
     }
 
@@ -168,7 +168,7 @@ public class AsyncOperationService extends AbstractService {
 
     public AsyncOperationStatus getOperationStatus(int requestId) {
         if(!operations.containsKey(requestId))
-            throw new VoldemortException("No operation with id " + requestId + " found");
+            throw new AsyncOperationNotFoundException(requestId);
 
         return operations.get(requestId).getStatus();
     }
@@ -187,7 +187,7 @@ public class AsyncOperationService extends AbstractService {
 
     public void stopOperation(int requestId) {
         if(!operations.containsKey(requestId))
-            throw new VoldemortException("No operation with id " + requestId + " found");
+            throw new AsyncOperationNotFoundException(requestId);
 
         operations.get(requestId).stop();
     }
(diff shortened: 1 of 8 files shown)


src/java/voldemort/server/protocol/admin/AsyncOperationService.java
```diff
@@ -112,7 +112,7 @@ public String getStatus(int id) {
         try {
             return getOperationStatus(id).toString();
         } catch(VoldemortException e) {
-            return "No operation with id " + id + " found";
+            return String.format(AsyncOperationNotFoundException.MESSAGE, id);
```
gaojieliu: Shouldn't it just return e.getMessage() ?

src/java/voldemort/utils/ExceptionUtils.java
```diff
@@ -55,4 +59,50 @@ public static String stackTraceToString(Throwable throwable) {
...
+     * @return true if it represents an {@link AsyncOperationNotFoundException}, false otherwise.
+     */
+    public static boolean isAsyncOpNotFound(Exception e) {
+        if (e instanceof AsyncOperationNotFoundException) {
+            // Updated servers should throw this.
+            return true;
+        } else if (e instanceof VoldemortException) {
+            // Legacy servers should throw a generic VoldEx identifiable with this regex
+            String exceptionMessage = e.getMessage();
+            String regex = String.format(AsyncOperationNotFoundException.MESSAGE, ".*");
```
gaojieliu: This could be a static string, and only be initialized one time.

src/java/voldemort/utils/ExceptionUtils.java
```diff
@@ -55,4 +59,50 @@ public static String stackTraceToString(Throwable throwable) {
...
+     * @return true if it represents an {@link StoreVersionAlreadyExistsException}, false otherwise.
+     */
+    public static boolean isStoreVersionAlreadyExists(Exception e) {
+        if (e instanceof StoreVersionAlreadyExistsException) {
+            // Updated servers should throw this.
+            return true;
+        } else if (e instanceof VoldemortException) {
+            // Legacy servers should throw a generic VoldEx identifiable with this regex
+            String exceptionMessage = e.getMessage();
+            String regex = String.format(StoreVersionAlreadyExistsException.MESSAGE, ".*");
```
gaojieliu: same here.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -146,6 +151,10 @@ public void invokeRollback(final String storeName, final long pushVersion) {
     public Map<Node, Response> invokeFetch(final String storeName,
                                            final String basePath,
                                            final long pushVersion) {
+        // This should result in 5 minutes max of wait time, for failure scenarios which are potentially recoverable.
```
gaojieliu: Is 5 mins enough for normal upgrade of one node? Or do you think it is worth creating a plugin config for it considering sometimes the deployment for remote colo could be slow.

src/java/voldemort/store/readonly/swapper/AdminStoreSwapper.java
```diff
@@ -146,6 +151,10 @@ public void invokeRollback(final String storeName, final long pushVersion) {
...
+        final int MAX_FETCH_ATTEMPT = 10;
+        final long WAIT_TIME_BETWEEN_ATTEMPTS = 30000;
```
gaojieliu: Where will those two variables be used? I could not see it in this pull request.

PR #463, opened 2016-11-24
Make nodes option working again for generate_cluster_xml.py

bitti:
The `nodes` option for `generate_cluster_xml.py` is confusing because it does nothing when the `file` option is provided, but the `file` option got mandatory about 3 years ago when the file exists check was added.

Therefore I propose this fix, or alternatively remove this option altogether (since it's not working since 3 years I guess nobody will miss it?)


FelixGV:
I merged it in. Thanks!

PR #464, opened 2016-11-28
Replace Unicode quotes in scripts with ASCII quotes

bitti:
Some issues I noticed during an IDEA analysis run. I don't know how these scripts could ever work, they seem to contain Unicode " (code point 0x201c) since the beginning (which can happen when copy and pasting from a blog for instance).

I couldn't test these scripts, but since they couldn't work anyway I'm wondering if anybody is using these. For the benefits of users outside of LinkedIn I would prefer to clean them up in order to make the Voldemort distribution less confusing. I could provide an alternative PR for that.

### DIFF
--- a/bin/rebalance-cluster-expansion.sh
+++ b/bin/rebalance-cluster-expansion.sh
@@ -60,7 +60,7 @@ dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 vold_home="$(dirname "$dir")"
 
 # Parse options
-while getopts “hc:s:i:o:” OPTION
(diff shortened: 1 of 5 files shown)


bitti:
Maybe it works because the `getopts` builtin just ignores these Unicode chars? But even then I think this should be fixed in order to avoid confusion.

FelixGV:
Merged. Thanks for all the contribs, @bitti !

bitti:
Thanks, appreciate it! I hope I can open some more substantial PRs soon

FelixGV:
Uh oh ;P ...

Just kidding, sounds great xD !

PR #465, opened 2016-11-29
Allow explicit setting of build.primary.replicas.only for BnP

bitti:
We needed to add this option since it's just assumed to be true when the server supports it, even though the store is build in the old format.

LinkedIn probably doesn't need this since I guess you're using the newer store format everywhere. Outsiders are not that fast though. We even didn't use the BnP method until recently and we only recently switched some stores to `build.primary.replicas.only` but by far not all. So this option would help us during the transition.

I don't know if there is a unit test covering this code path, but I can surely add a test if you point me to some.

### DIFF
--- a/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
+++ b/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/azkaban/VoldemortBuildAndPushJob.java
@@ -521,18 +521,21 @@ public class VoldemortBuildAndPushJob extends AbstractJob {
 
         // 3. Get "build.primary.replicas.only" setting
 
-        Map<String, String> expectedConfig = Maps.newHashMap();
-        expectedConfig.put(VoldemortConfig.READONLY_BUILD_PRIMARY_REPLICAS_ONLY, Boolean.toString(true));
-        this.buildPrimaryReplicasOnly = true;
-        for (String clusterUrl: clusterURLs) {
-            VAdminProto.GetHighAvailabilitySettingsResponse serverSettings = haSettingsPerCluster.get(clusterUrl);
-            int maxNodeFailuresForCluster = 0;
-            if (serverSettings != null) {
-                maxNodeFailuresForCluster = serverSettings.getMaxNodeFailure();
-            }
-            if (!adminClientPerCluster.get(clusterUrl).metadataMgmtOps.validateServerConfig(expectedConfig, maxNodeFailuresForCluster)) {
-                log.info("'" + BUILD_PRIMARY_REPLICAS_ONLY + "' is not supported on this destination cluster: " + clusterUrl);
-                this.buildPrimaryReplicasOnly = false;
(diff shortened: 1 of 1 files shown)


FelixGV:
Hmm...

So if I understand correctly, the concern is that `build.primary.replicas.only` (or BPRO) does not work in conjunction with `save.keys=false`.

I think this is a bug we should fix, but I would like to avoid giving the user enough rope to hang themselves with. Instead of adding a new config flag, which needs to be set right, I would instead lean towards making BnP "do the right thing" directly.

How about this: in the constructor, if `save.keys` is false, then BPRO should be set to false without consulting the servers. Otherwise, we let the current code run its course (i.e.: ask the servers if they support BPRO; if they do, rely on their server-side config to make the call, and if they don't disable BPRO).

What do you think about the above?

Regarding tests: Unfortunately, BnP in general is poorly tested. There are a few integration and unit tests, though not many, and there aren't any unit tests specifically for BPRO.

-F

bitti:
I don't understand why it has to be tangled like that. BPRO implies `save.keys` but not the other way around. We even have tested `save.keys` for some stores without enabling BPRO, so in this case we would have the same problem again (that is that it's just assumed). And why should it give the user rope for hanging? The default is still true so no one currently using it has to mind it, nothing changes for them. In fact it only needs to be explicitly set when it's set to `false`, so for usecases like ours when one should know why.

Also when you set `save.keys` to false but BPRO to true you just get an corresponding error, so it fails early and the fix is clear for the user. But when we were setting a node directory for `build.output.dir` (because the we figured the parent directory wasn't working for some reason, because we didn't know at that time that BPRO is implicitly enabled) the fetcher happily went on until we finally noticed that it just replicated the same partitions meant for 1 node to all nodes. That is a real rope to hang oneself.

FelixGV:
Well, I may be missing something, but I don't see a check which fails fast in the case where `save.keys == false && BPRO == true`. Do we have such check? If so, where? This particular case is the one where I want to avoid giving the user enough rope to hang themselves with.

If we do have a fail fast mechanism already, then that's fine. If we don't, then we should either add a fail fast mechanism, or add some reasonable default.

bitti:
It's checked during in the [store builder](https://github.com/voldemort/voldemort/blob/ea37ef67fa7724180608510c6d4237167b78dd63/contrib/hadoop-store-builder/src/java/voldemort/store/readonly/mr/AbstractStoreBuilderConfigurable.java#L130). But I agree that another check in Push would be helpful, because one doesn't have to use the push configuration also for building so they can differ if misconfigured (in fact we are still building by calling the `HadoopStoreBuilder` class directly and using BnP only for push -- did I mention we are slow to migrate?). The fact that it still started fetching as a BPRO store when it was not even though `save.keys` was set to false is concerning. But I still would like to be able to set the `build.primary.replicas.only` option explicitly, because you can have `save.keys` enabled without BPRO.

So I try come up with another patch which does some checks to see if the configuration is consistent.

If you want to avoid user confusion, the best way would actually be to get rid of the BPRO and `save.keys` option. I don't see any advantage in not using BPRO (other than not having to migrate legacy stores), so it's just a confusing option to have. I'm clear that this can only happen in a  major release (Voldemort 2.0?), but in the meantime you have to give your users some way to get there.

PR #467, opened 2017-01-06
Fix HA issue when build.replica.factor is 1

gaojieliu:
Added build.replica.factor check when decide whether Voldemort should fail the
data push when data fetches fail in some nodes in HA mode.

@FelixGV


wt:
LGTM.

gaojieliu:
Merged.

PR #471, opened 2017-02-27
New RECOVERY mode to partially automate node repair following BnP HA.

FelixGV:
In the new RECOVERY mode, the server will deny online requests from
clients (just like in OFFLINE mode) but will allow new BnP jobs to
push into the node. After starting the server in this mode, the
operator can kick off a data restoration job to recover the rest of
the missing stores. Meanwhile, the server will periodically check
if any of its symlinks have changed, and if so will load the newly
restored stores as the active version. Finally, when no more stores
are disabled, the server automatically transitions to the ONLINE
mode and resume serving client requests.

The benefits of this new mode are:

1) Less manual work for the operator.
2) No need for a planned BnP outage during the node recovery process.

FYI: This is untested. It barely compiles. Hot off the IDE. Handle with care.

### DIFF
--- a/src/java/voldemort/server/VoldemortServer.java
+++ b/src/java/voldemort/server/VoldemortServer.java
@@ -47,6 +47,7 @@ import voldemort.server.niosocket.NioSocketService;
 import voldemort.server.protocol.ClientRequestHandlerFactory;
 import voldemort.server.protocol.RequestHandlerFactory;
 import voldemort.server.protocol.SocketRequestHandlerFactory;
+import voldemort.server.protocol.admin.AsyncOperation;
 import voldemort.server.protocol.admin.AsyncOperationService;
 import voldemort.server.rebalance.Rebalancer;
 import voldemort.server.rebalance.RebalancerService;
@@ -469,11 +470,8 @@ public class VoldemortServer extends AbstractService {
         logger.info("Starting " + basicServices.size() + " services.");
         long start = System.currentTimeMillis();
 
-        boolean goOnline;
-        if (getMetadataStore().getServerStateUnlocked() == MetadataStore.VoldemortState.OFFLINE_SERVER)
-            goOnline = false;
-        else
-            goOnline = true;
+        MetadataStore.VoldemortState serverState = getMetadataStore().getServerStateUnlocked();
+        boolean goOnline = serverState.enableOnlineServices;
 
         for(VoldemortService service: basicServices) {
             try {
@@ -487,7 +485,13 @@ public class VoldemortServer extends AbstractService {
         if (goOnline) {
             startOnlineServices();
         } else {
-            goOffline();
+            switch (serverState) {
+                case RECOVERY: {
+                    goIntoRecoveryMode();
+                    break;
+                }
+                default: goOffline();
+            }
         }
         long end = System.currentTimeMillis();
         logger.info("Startup completed in " + (end - start) + " ms.");
@@ -596,7 +600,7 @@ public class VoldemortServer extends AbstractService {
     }
 
     public void goOffline() {
-        getMetadataStore().setOfflineState(true);
+        getMetadataStore().setState(MetadataStore.VoldemortState.OFFLINE_SERVER);
         stopOnlineServices();
     }
 
@@ -604,7 +608,7 @@ public class VoldemortServer extends AbstractService {
         ReadOnlyStoreStatusValidation validation = validateReadOnlyStoreStatusBeforeGoingOnline();
 
         if (validation.readyToGoOnline) {
-            getMetadataStore().setOfflineState(false);
+            getMetadataStore().setState(MetadataStore.VoldemortState.NORMAL_SERVER);
             createOnlineServices();
             startOnlineServices();
         }
@@ -612,6 +616,35 @@ public class VoldemortServer extends AbstractService {
         if (validation.e != null) {
             throw new VoldemortException("Problem while going online!", validation.e);
         }
+    }
+
+    public void goIntoRecoveryMode() {
+        getMetadataStore().setState(MetadataStore.VoldemortState.RECOVERY);
+
+        // Spin up an async task to check RO stores periodically...
+        int asyncOpId = asyncService.getUniqueRequestId();
+        asyncService.submitOperation(asyncOpId, new AsyncOperation(asyncOpId, "RecoveryTask") {
+            private boolean keepGoing = true;
+
+            @Override
+            public void operate() throws Exception {
+                while (getMetadataStore().getServerStateLocked() == MetadataStore.VoldemortState.RECOVERY && keepGoing) {
+                    try {
+                        goOnline();
+                    } catch (VoldemortException e) {
+                        long sleepTime = getVoldemortConfig().getHighAvailabilityRecoveryAttemptIntervalSec();
+                        logger.info("Recovery task tried going online, but the server is not ready yet. " +
+                                "Will sleep " + sleepTime + " seconds before next attempt.");
+                        Thread.sleep(sleepTime);
+                    }
+                }
+            }
+
+            @Override
+            public void stop() {
+                keepGoing = false;
+            }
+        });
     }
 
     private class ReadOnlyStoreStatusValidation {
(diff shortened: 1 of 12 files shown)


src/java/voldemort/store/metadata/MetadataStore.java
```diff
@@ -123,16 +123,37 @@
...
+     put(READONLY_FETCH_ENABLED_KEY, false);
+     initCache(READONLY_FETCH_ENABLED_KEY);
+
      */
-    public static enum VoldemortState {
-        NORMAL_SERVER,
-        OFFLINE_SERVER,
-        REBALANCING_MASTER_SERVER
+    public enum VoldemortState {
+        /** The default state. */
```
gnb: This is a great way of encapsulating the semantics of the states.  Nice!

src/java/voldemort/store/metadata/MetadataStore.java
```diff
@@ -947,58 +968,28 @@ public void deleteRebalancingState(RebalanceTaskInfo stealInfo) {
...
-                } else if(currentState.equals(VoldemortState.OFFLINE_SERVER.toString())) {
-                    put(SERVER_STATE_KEY, VoldemortState.NORMAL_SERVER);
-                    initCache(SERVER_STATE_KEY);
-                    put(SLOP_STREAMING_ENABLED_KEY, true);
-                    initCache(SLOP_STREAMING_ENABLED_KEY);
-                    put(PARTITION_STREAMING_ENABLED_KEY, true);
-                    initCache(PARTITION_STREAMING_ENABLED_KEY);
-                    put(READONLY_FETCH_ENABLED_KEY, true);
-                    initCache(READONLY_FETCH_ENABLED_KEY);
-                    init();
```
gnb: These two lines do not have an equivalent in the new code.

src/java/voldemort/store/metadata/MetadataStore.java
```diff
@@ -947,58 +968,28 @@ public void deleteRebalancingState(RebalanceTaskInfo stealInfo) {
...
-                    initCache(SLOP_STREAMING_ENABLED_KEY);
-                    put(PARTITION_STREAMING_ENABLED_KEY, false);
-                    initCache(PARTITION_STREAMING_ENABLED_KEY);
-                    put(READONLY_FETCH_ENABLED_KEY, false);
-                    initCache(READONLY_FETCH_ENABLED_KEY);
-                } else if(currentState.equals(VoldemortState.OFFLINE_SERVER.toString())) {
-                    logger.warn("Already in OFFLINE_SERVER state.");
-                    return;
-                } else {
-                    logger.error("Cannot enter OFFLINE_SERVER state from " + currentState);
```
gnb: The new code does not attempt to prevent some transitions.  Is this a deliberate policy?

src/java/voldemort/store/metadata/MetadataStore.java
```diff
@@ -947,58 +968,28 @@ public void deleteRebalancingState(RebalanceTaskInfo stealInfo) {
...
+
+            // Set metadata properties corresponding to newState:
+            put(SERVER_STATE_KEY, newState);
+            initCache(SERVER_STATE_KEY);
+            put(SLOP_STREAMING_ENABLED_KEY, newState.enableSlopStreaming);
+            initCache(SLOP_STREAMING_ENABLED_KEY);
+            put(PARTITION_STREAMING_ENABLED_KEY, newState.enablePartitionStreaming);
+            initCache(PARTITION_STREAMING_ENABLED_KEY);
+            put(READONLY_FETCH_ENABLED_KEY, newState.enableReadOnlyFetches);
+            initCache(READONLY_FETCH_ENABLED_KEY);
```
gnb: Nice.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -176,6 +178,27 @@ public String getChunkIdToNumChunks() {
         return builder.toString();
     }
 
+    public void reSyncFileSystem() {
+        fileModificationLock.readLock().lock();
+        File latestSymLink = new File(latestDirSymLinkPath);
+        boolean needToResync = !Utils.symlinkExists(ReadOnlyUtils.getCurrentVersion(storeDir), latestSymLink);
```
gnb: Are you going to leak the `fileModificationLock` if `symlinkExists()` throws an exception?

src/java/voldemort/utils/Utils.java
```diff
@@ -148,45 +148,60 @@ public static void rm(Iterable<File> files) {
...
-        File symLink = new File(symLinkPath);
-
+     * Checks if a symlink already exists and already points to the intended destination
+     *
+     * @param file The file the symbolic link should point to
+     * @param symLink The symbolic link which should already exist
+     * @return true if the symlink exists and points to the intended destination, false otherwise.
+     * @throws VoldemortException if the destination file does not exist or IOException occurs.
+     */
+    public static boolean symlinkExists(File file, File symLink) {
```
gnb: If I understand this function, it has four possible results:

1. symlink exists and has the expected value: return true, caller does nothing and does not need to keep trying
2. symlink exists and has a different valid value: return false, caller tries to resync from the filesystem
3. symlink doesn't exist: throw exception, caller does nothing and (sensibly) keeps trying
4. actually exceptional condition (parent directory missing, permissions issues): throw exception, caller does nothing and (foolishly) keeps trying

My problem is that case 3 is not sufficiently distinguished.  It's not an exceptional condition.  During recovery it's a normal and expected condition that can last for hours but will eventually go away.  Lumping together cases 3 and 4 into an undistinguished exception that results in the caller blindly retrying without logging, will make it harder to diagnose problems like filesystem corruption or permissions issues.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -176,6 +178,27 @@ public String getChunkIdToNumChunks() {
...
+        fileModificationLock.readLock().unlock();
+
+        if (needToResync) {
+            logger.info("The internal state is inconsistent with the filesystem's state. Resyncing.");
```
gnb: This message is both vague and ominous.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
```diff
@@ -176,6 +178,27 @@ public String getChunkIdToNumChunks() {
...
+            File newLatestVersion = null;
+            try {
+                newLatestVersion = latestSymLink.getCanonicalFile();
```
gnb: We just got this in `symlinkExists()` under the lock.  Could the `symlinkExists()` API be adjusted to make this more efficient and safe?

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -614,6 +618,35 @@ public void goOnline() {
         }
     }
 
+    public void goIntoRecoveryMode() {
+        getMetadataStore().setState(MetadataStore.VoldemortState.RECOVERY);
```
gnb: It might be useful to have a log message like "Entering recovery mode" here.

src/java/voldemort/server/VoldemortServer.java
```diff
@@ -614,6 +618,35 @@ public void goOnline() {
...
+                while (getMetadataStore().getServerStateLocked() == MetadataStore.VoldemortState.RECOVERY && keepGoing) {
+                    try {
+                        goOnline();
+                    } catch (VoldemortException e) {
+                        long sleepTime = getVoldemortConfig().getHighAvailabilityRecoveryAttemptIntervalSec();
+                        logger.info("Recovery task tried going online, but the server is not ready yet. " +
+                                "Will sleep " + sleepTime + " seconds before next attempt.");
+                        Thread.sleep(sleepTime);
+                    }
+                }
```
gnb: it might be useful to have a log message like "Recovery mode finished" here.

src/java/voldemort/store/readonly/ReadOnlyStorageEngine.java
gnb: Actually it would probably be useful to log the current and apparent new values of the version here.

PR #473, opened 2017-03-14
produce a fatjar that works in our testsuite. convert to avro 1.8.1

christophsturm:


christophsturm:
oops sorry i wanted to make this pr in our fork and not here :)

bitti:
Why every clone has to fix this anew? Can't we just merge https://github.com/voldemort/voldemort/pull/470? (Avro 1.7.7 is not quite 1.8.1, but I suggest that the difference is minor regarding the API changes). 

I understand that similar PR's were rejected years ago, but since then the premise changed with the availability of shadowing.

PR #486, opened 2017-08-08
Fixed the issue with too long command line in Windows 10 + updated the readme file to reflect proper build instructions

### LINKED ISSUE #485: Unable to build and run Voldemort on Windows 10
I've followed the instructions in the README.md file, but I wasn't able to run the server under Windows 10. When executing bin\voldemort-server.bat config/single_node_cluster, it's missing the Log4J library and when I build the project with "gradlew clean build -x test" instead of "gradlew clean jar" then I get: "Input line is too long".

zsoltkatona:
The problem under Windows 10 was that:
1. Using the build instructions in the readme file (i.e. gradlew clean jar) didn't result in the lib folder created and the dependent jars copied into it therefore at startup the first library to be loaded (Log4J) could not be loaded and caused the application to stop.
2. After proper build (i.e. gradlew clean build -x test), the build succeeded, the lib folder and dependent jars were there, but during the execution of the bat file the concatenation of the actual command to run resulted in a too long command line at some point and the execution halted. It was unnecessary however to do the concatenation of the absolute paths in the lib folder, Java accepts wildcard in classpath definitions, so I just replaced that logic with appending a single "lib/*" to the classpath.

This fixes #485.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ It is used at LinkedIn by numerous critical services powering a large portion of
 cd ~/workspace
 git clone https://github.com/voldemort/voldemort.git
 cd voldemort
-./gradlew clean jar
+./gradlew clean build -x test
 ```
 
 ### Start Server ###
--- a/bin/run-class.bat
+++ b/bin/run-class.bat
@@ -29,9 +29,9 @@ SET CLASSPATH=.
 
 set VOLDEMORT_CONFIG_DIR=%1%/config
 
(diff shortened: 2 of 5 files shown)


bitti:
It was necessary when the file was created, because Voldemort still was compiled with Java 6 then, which doesn't support wildcards in classpaths. But since the upgrade of Voldemort to Java 7 this is a welcome simplification I think.

FelixGV:
I merged this in. Thanks!

PR #488, opened 2017-08-30
The RO server is now returning old async op IDs idempotently.

FelixGV:
Previously, the BnP job could retry a fetch if it had connectivity
issues midway through, and the server would reject the new fetch
if the previous one was already going on. Now, it will instead
return the previous async op ID if the new fetch request is for the
same version as previously requested.


FelixGV:
This is merged in already.

